PR #2, opened 2018-12-06
wip hub 2

tlwr:
@tlwr @sakisv

### DIFF
--- a/terraform/modules/hub/hub_config.tf
+++ b/terraform/modules/hub/hub_config.tf
@@ -6,7 +6,8 @@ module "config_ecs_asg" {
   cluster             = "config"
   vpc_id              = "${aws_vpc.hub.id}"
   instance_subnets    = ["${aws_subnet.internal.*.id}"]
-  number_of_instances = "${var.number_of_availability_zones}"
+  # number_of_instances = "${var.number_of_availability_zones}"
+  number_of_instances = 0
   domain              = "${var.domain}"
 
   additional_instance_security_group_ids = [
@@ -29,14 +30,15 @@ module "config" {
   cluster               = "config"
   domain                = "${var.domain}"
   vpc_id                = "${aws_vpc.hub.id}"
-  task_subnets               = ["${aws_subnet.internal.*.id}"]
+  task_subnets          = ["${aws_subnet.internal.*.id}"]
   task_definition       = "${data.template_file.config_task_def.rendered}"
   container_name        = "stub"
   container_port        = "8080"
   number_of_tasks       = 1
   health_check_protocol = "HTTP"
+  tools_account_id      = "${var.tools_account_id}"
 
   additional_task_security_group_ids = [
-    "${aws_security_group.egress_via_proxy.id}"
+    "${aws_security_group.egress_via_proxy.id}",
   ]
 }
--- a/terraform/modules/hub/hub_frontend.tf
+++ b/terraform/modules/hub/hub_frontend.tf
@@ -15,9 +15,45 @@ resource "aws_security_group" "frontend_task" {
 }
 
 data "template_file" "frontend_task_def" {
-  template = "${file("${path.module}/files/tasks/stub.json")}"
+  template = "${file("${path.module}/files/tasks/frontend.json")}"
 
   vars {
-    app = "frontend"
+    app           = "frontend"
+    image_and_tag = "${local.tools_account_ecr_url_prefix}-verify-frontend:latest"
+    domain        = "${var.domain}"
+  }
+}
+
+module "frontend_ecs_roles" {
+  source = "modules/ecs_iam_role_pair"
+
+  deployment       = "${var.deployment}"
+  tools_account_id = "${var.tools_account_id}"
+  service_name     = "frontend"
+  image_name       = "verify-frontend"
+}
+
+resource "aws_ecs_task_definition" "frontend" {
+  family                = "${var.deployment}-frontend"
+  container_definitions = "${data.template_file.frontend_task_def.rendered}"
+  network_mode          = "awsvpc"
+  execution_role_arn    = "${module.frontend_ecs_roles.execution_role_arn}"
+}
+
+resource "aws_ecs_service" "frontend" {
+  name            = "${var.deployment}-frontend"
+  cluster         = "${aws_ecs_cluster.ingress.id}"
+  task_definition = "${aws_ecs_task_definition.frontend.arn}"
+  desired_count   = 1
+
+  load_balancer {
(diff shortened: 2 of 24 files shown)


terraform/modules/hub/hub_config.tf
```diff
@@ -6,7 +6,8 @@ module "config_ecs_asg" {
   cluster             = "config"
   vpc_id              = "${aws_vpc.hub.id}"
   instance_subnets    = ["${aws_subnet.internal.*.id}"]
-  number_of_instances = "${var.number_of_availability_zones}"
+  # number_of_instances = "${var.number_of_availability_zones}"
+  number_of_instances = 0
```
issyl0: Why is this 0? Do we still need to make `config` the app exist?

terraform/modules/hub/hub_frontend.tf
```diff
@@ -15,9 +15,45 @@ resource "aws_security_group" "frontend_task" {
...
+  deployment       = "${var.deployment}"
+  tools_account_id = "${var.tools_account_id}"
+  service_name     = "frontend"
+  image_name       = "verify-frontend"
+}
+
+resource "aws_ecs_task_definition" "frontend" {
+  family                = "${var.deployment}-frontend"
+  container_definitions = "${data.template_file.frontend_task_def.rendered}"
+  network_mode          = "awsvpc"
```
issyl0: We're not sure what this does?

terraform/modules/hub/hub_metadata.tf
```diff
@@ -41,6 +42,14 @@ data "template_file" "metadata_task_def" {
   }
 }
 
+module "metadata_ecs_roles" {
+  source = "modules/ecs_iam_role_pair"
+
+  deployment       = "${var.deployment}"
+  service_name     = "metadata"
+  tools_account_id = "${var.tools_account_id}"
```
issyl0: Is this for Concourse to be able to deploy?

terraform/modules/hub/hub_policy.tf
```diff
@@ -6,7 +6,8 @@ module "policy_ecs_asg" {
   cluster             = "policy"
   vpc_id              = "${aws_vpc.hub.id}"
   instance_subnets    = ["${aws_subnet.internal.*.id}"]
-  number_of_instances = "${var.number_of_availability_zones}"
+  # number_of_instances = "${var.number_of_availability_zones}"
+  number_of_instances = 0
```
issyl0: How does `number_of_instances` relate to line 37, `number_of_tasks`? A task runs on an instance, so if there's 1 task but 0 instances then it won't work?

terraform/modules/hub/hub_saml_engine.tf
```diff
@@ -6,7 +6,8 @@ module "saml_engine_ecs_asg" {
   cluster             = "saml-engine"
   vpc_id              = "${aws_vpc.hub.id}"
   instance_subnets    = ["${aws_subnet.internal.*.id}"]
-  number_of_instances = "${var.number_of_availability_zones}"
+  # number_of_instances = "${var.number_of_availability_zones}"
+  number_of_instances = 0
```
issyl0: As above.

terraform/modules/hub/ingress.tf
```diff
@@ -57,6 +59,23 @@ resource "aws_lb_target_group" "ingress_metadata" {
...
+  vpc_id               = "${aws_vpc.hub.id}"
+  target_type          = "ip"
+  deregistration_delay = 60
+
+  health_check {
+    path     = "/"
+    protocol = "HTTP"
+    interval = 10
+    timeout  = 5
+    matcher  = "200,301"
```
issyl0: Why are there two possible status codes?

terraform/modules/hub/vpc.tf
```diff
@@ -33,7 +33,8 @@ resource "aws_vpc_endpoint" "s3" {
             ],
             "Resource": [
                 "arn:aws:s3:::govukverify-eidas-metadata-aggregator-${var.deployment}-a/*",
-                "arn:aws:s3:::govukverify-eidas-metadata-aggregator-${var.deployment}-a"
+                "arn:aws:s3:::govukverify-eidas-metadata-aggregator-${var.deployment}-a",
```
issyl0: Why `-a` and not `e`? Or are we getting things from/to to the Old World?

tlwr: this is an existing bucket
we're going to change how this works but for now we just proxy to the existing bucket

https://govukverify-eidas-metadata-aggregator-joint-a.s3-eu-west-2.amazonaws.com/

terraform/modules/hub/ingress.tf
tlwr: RAILS_ENV=production => 301

RAILS_ENV=development => 200

terraform/modules/hub/hub_metadata.tf
tlwr: It is so we can get images from the tools ECR

terraform/modules/hub/hub_frontend.tf
tlwr: I would read up on awsvpc networking, we'll strip it out from all of the other apps except frontend/metadata/analytics

All of the other boxes like config/policy/saml-* are one type of app per box so we can just use the box security groups

terraform/modules/hub/hub_config.tf
tlwr: yeah we do but we're not deploying to it so until then save $

terraform/modules/hub/hub_policy.tf
tlwr: in this case it just wont try and schedule a task as there is no place to schedule it

terraform/modules/hub/modules/ecs_iam_role_pair/task_role.tf
```diff
@@ -0,0 +1,21 @@
...
+      {
+        "Effect": "Allow",
+        "Principal": {
+          "Service": "ecs-tasks.amazonaws.com"
+        },
+        "Action": "sts:AssumeRole"
+      }
+    ]
+  }
+  EOF
```
sakisv: I assume there are no mentions of specific permissions for the task-role because for now we have only frontend which doesn't need it, right?

terraform/modules/hub/hub_frontend.tf
sakisv: Right, so when the docs say

> The task networking features provided by the awsvpc network mode give Amazon ECS tasks the same networking properties as Amazon EC2 instances.

They don't mean it will have the same properties as your host EC2 instance, rather that the containers networking will be treated as if itself is an EC2 instance.

PR #6, opened 2018-12-13
Enable and add task definition for saml-soap-proxy

sakisv:
@bartlabno @sakisv

### DIFF
--- a/terraform/modules/hub/hub_saml_soap_proxy.tf
+++ b/terraform/modules/hub/hub_saml_soap_proxy.tf
@@ -43,3 +43,12 @@ module "saml_soap_proxy" {
   certificate_arn            = "${data.aws_acm_certificate.wildcard.arn}"
   image_name                 = "verify-saml-soap-proxy"
 }
+
+module "saml_soap_proxy_can_connect_to_config" {
+  source = "modules/microservice_connection"
+
+  source_sg_id      = "${module.saml_soap_proxy_ecs_asg.instance_sg_id}"
+  destination_sg_id = "${module.config.lb_sg_id}"
+
+  port = 443
+}
--- a/terraform/modules/hub/hub_saml_soap_proxy.tf
+++ b/terraform/modules/hub/hub_saml_soap_proxy.tf
@@ -1,14 +1,12 @@
 module "saml_soap_proxy_ecs_asg" {
   source = "modules/ecs_asg"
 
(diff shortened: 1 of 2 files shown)


terraform/modules/hub/egress_proxy.tf
```diff
@@ -50,6 +50,7 @@ locals {
     "auth\\.docker\\.io",                                                      # Docker Hub
     "production\\.cloudflare\\.docker\\.com",                                  # Docker Hub
     "www\\.e\\.${var.deployment}\\.signin\\.service\\.gov\\.uk",               # Metadata
+    "config.e\\.${var.deployment}\\.signin\\.service\\.gov\\.uk",              # Confi
```
issyl0: Typo in the comment.

terraform/modules/hub/hub_saml_soap_proxy.tf
```diff
@@ -17,27 +15,31 @@ module "saml_soap_proxy_ecs_asg" {
...
+  task_definition            = "${data.template_file.saml_soap_proxy_task_def.rendered}"
+  container_name             = "saml-soap-proxy"
+  container_port             = "8080"
+  number_of_tasks            = 1
+  health_check_protocol      = "HTTP"
+  health_check_path          = "/saml-soap-proxy/matching-service-health-check"
+  tools_account_id           = "${var.tools_account_id}"
   instance_security_group_id = "${module.saml_soap_proxy_ecs_asg.instance_sg_id}"
   certificate_arn            = "${data.aws_acm_certificate.wildcard.arn}"
+  image_name                 = "verify-saml-soap-proxy"
```
issyl0: What's the difference between `image_name` and `image_and_tag` in the variables?

terraform/modules/hub/egress_proxy.tf
sakisv: Good catch, but I will need to remove this line anyway since the apps are expected to talk to the config internally without having to go through the egress-proxy.

terraform/modules/hub/hub_saml_soap_proxy.tf
tlwr: yeah we could use image_name and image_tag

and combine them in the right places

right now image_and_tag goes into the task definition

and image_name is used for permissions to pull the image

tlwr:
matching service health check, there should be a service status endpoint

we can look at this tomorrow

PR #8, opened 2018-12-20
SAML Engine changes

adityapahuja:
Please see commits for further details.

### DIFF
--- a/terraform/modules/hub/files/tasks/hub-saml-engine.json
+++ b/terraform/modules/hub/files/tasks/hub-saml-engine.json
@@ -11,7 +11,21 @@
         "hostPort": 8080
       }
     ],
-    "entryPoint": ["/tmp/saml-engine-entrypoint.sh"],
+    "secrets": [
+      {
+        "name": "HUB_SIGNING_PRIVATE_KEY",
+        "valueFrom": "arn:aws:ssm:${region}:${account_id}:parameter/${deployment}-hub-signing-private-key"
+      },
+      {
+        "name": "PRIMARY_HUB_ENCRYPTION_PRIVATE_KEY",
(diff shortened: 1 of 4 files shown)


terraform/modules/hub/files/tasks/hub-saml-engine.json
```diff
@@ -11,7 +11,21 @@
...
+        "name": "HUB_SIGNING_PRIVATE_KEY",
+        "valueFrom": "arn:aws:ssm:${region}:${account_id}:parameter/${deployment}-hub-signing-private-key"
+      },
+      {
+        "name": "PRIMARY_HUB_ENCRYPTION_PRIVATE_KEY",
+        "valueFrom": "arn:aws:ssm:${region}:${account_id}:parameter/${deployment}-primary-hub-encryption-private-key"
+      },
+      {
+        "name": "SECONDARY_HUB_ENCRYPTION_PRIVATE_KEY",
+        "valueFrom": "arn:aws:ssm:${region}:${account_id}:parameter/${deployment}-secondary-hub-encryption-private-key"
```
sakisv: I know these are encrypted with the our KMS CMK, but why are we not using secrets here?

terraform/modules/hub/hub_saml_engine.tf
```diff
@@ -64,3 +66,32 @@ module "saml_engine_can_connect_to_saml_soap_proxy" {
...
+  name = "${var.deployment}-saml-engine-parameter-execution"
+
+  policy = <<-EOF
+  {
+    "Version": "2012-10-17",
+    "Statement": [{
+      "Effect": "Allow",
+      "Action": [
+        "ssm:GetParameters",
+        "secretsmanager:GetSecretValue",
```
sakisv: If we're only using parameters is this action still needed?

PR #16, opened 2019-01-08
Add nginx in front of applications

adityapahuja:
This pull request adds nginx in front of all applications to act as a SSL termination and is used for encrypting communications between Application Load Balancer (ALB) and nginx over the network since they are not running on the same server or EC2 instance.

### DIFF
--- a/terraform/modules/hub/hub_config.tf
+++ b/terraform/modules/hub/hub_config.tf
@@ -18,33 +18,47 @@ module "config_ecs_asg" {
   logit_elasticsearch_url = "${var.logit_elasticsearch_url}"
 }
 
+locals {
+  config_location_blocks = <<-LOCATIONS
+  location / {
+    proxy_pass http://config:8080;
+    proxy_set_header Host config.${local.root_domain};
+  }
+  LOCATIONS
+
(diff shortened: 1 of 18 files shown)


terraform/modules/hub/hub_config.tf
```diff
@@ -18,14 +18,31 @@ module "config_ecs_asg" {
   logit_elasticsearch_url = "${var.logit_elasticsearch_url}"
 }
 
+locals {
+  config_location_blocks = <<-LOCATIONS
+  location = /healthcheck {
```
tlwr: can we remove the healthcheck and use /service-status for config?

tlwr: or rather /service status is used but this contains a healthcheck location for no reason

adityapahuja: Thanks for spotting. It looks like My rebase error has brought the old code back. I will remove it.

tlwr: :)

adityapahuja: I have resolved the rebase error. The old code is no longer present.

PR #18, opened 2019-01-14
Run the same number of Prometheis as we have AZs

issyl0:
- We want to have highly available Prometheis, so that means > 1.
- For non-prod environments, we run in 2 AZs rather than three.
- Match the number of Prometheis to the number of AZs, so we have 2 or 3
  Prometheis per environment.

Co-authored-by: Matthew Cullum <matthew.cullum@digital.cabinet-office.gov.uk>


issyl0:
`Plan` and `apply` successful.

PR #24, opened 2019-01-15
AUTO110: Update Prometheus to scrape metrics from all apps

adityapahuja:
This pull request contains the following changes.

- Update nginx to allow requests to go to prometheus metrics endpoints in all Hub applications (Config, SAML Engine, SAML Proxy and SAML SOAP Proxy).
- Update the applications' task definitions files to include a specified PROMETHEUS_METRICS_PATH environment variable to request them to use the specified Prometheus metrics endpoints.

Prometheus server currently uses https (port number: 8443) to scrape metrics data from policy application in Joint environment.

Author: @adityapahuja

### DIFF
--- a/terraform/modules/hub/files/tasks/hub-config.json
+++ b/terraform/modules/hub/files/tasks/hub-config.json
@@ -36,20 +36,31 @@
       "server",
       "/tmp/config.yml"
     ],
-    "environment": [{
-      "Name": "CONFIG_DATA_PATH",
-      "Value": "/tmp/federation/configuration/config-service-data/${deployment}"
-    }, {
-      "Name": "RP_TRUSTSTORE_PASSWORD",
-      "Value": "${truststore_password}"
-    }, {
-      "Name": "CLIENT_TRUSTSTORE_PASSWORD",
-      "Value": "${truststore_password}"
-    }, {
-      "Name": "DEPLOYMENT",
-      "Value": "${deployment}"
-    }, {
-      "Name": "JAVA_OPTS",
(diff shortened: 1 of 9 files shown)


terraform/modules/hub/files/tasks/hub-config.json
```diff
@@ -36,20 +36,31 @@
...
+      {
+        "Name": "DEPLOYMENT",
+        "Value": "${deployment}"
+      },
+      {
+        "Name": "JAVA_OPTS",
+        "Value": "-Dservice.name=config -Xms3000m -Xmx3200m -XX:+HeapDumpOnOutOfMemoryError -Dhttp.proxyHost=\"egress-proxy.${domain}\" -Dhttp.proxyPort=\"8080\" -Dhttps.proxyHost=\"egress-proxy.${domain}\" -Dhttps.proxyPort=\"8080\" -Dhttp.nonProxyHosts=\"\" -Dnetworkaddress.cache.ttl=5 -Dnetworkaddress.cache.negative.ttl=5",
+      },
+      {
+        "Name": "PROMETHEUS_METRICS_PATH",
```
philandstuff: What's this environment variable for? Who consumes it? Note that `engineering.reliability.gds.metrics.config.Configuration` (the only place I could find which uses this envvar) isn't used by PrometheusBundle (only MetricsBundle which is the "old and busted" way of doing things).

I think we can get rid of all of these.

adityapahuja: Hub application consumes it.

From readme file from https://github.com/alphagov/gds_metrics_dropwizard#how-to-build-your-project,  it says 

By default, metrics will be exposed at the path /metrics. You can change this with an environment variable like this:

```
PROMETHEUS_METRICS_PATH=/metrics
```
Dropwizard already has this endpoint `/metrics` on the admin port. That's why I added the environment variable to use a different endpoint `/prometheus/metrics` for Prometheus metrics.

adityapahuja: I see Prometheus Bundle and Metrics Bundle are different. Could you please update the readme file in gds_metrics_dropwizard repo about this change.

PR #28, opened 2019-01-16
BAU: Prometheus enhancement changes

adityapahuja:
This pull request updates Prometheus server to scrape metrics from specified list of applications which expose their endpoint (/prometheus/metrics) and ignore other applications or EC2 boxes which do not expose the endpoint. It also updates Prometheus script to read Prometheus configuration YAML file instead of having Prometheus configuration details hardcoded in it.


tlwr:
🚀

PR #29, opened 2019-01-16
BAU: Update Node Exporter to ignore docker directory

adityapahuja:
Prometheus Node Exporter does not have permission to read /var/lib/docker directory and it generates permission denied errors regularly in its log file. This commit updates Prometheus Node Exporter not to access /var/lib/docker directory. This will eliminate permission denied errors from the log file. It also updates cloud-init.sh to restart prometheus-node-exporter to pick up config changes.

Co-authored-by: Isabell (Issy) Long <me@issyl0.co.uk>

### DIFF
--- a/terraform/modules/hub/modules/ecs_asg/files/cloud-init.sh
+++ b/terraform/modules/hub/modules/ecs_asg/files/cloud-init.sh
@@ -152,5 +152,6 @@ docker run \
   amazon/amazon-ecs-agent:v1.23.0
 
 apt-get install --yes prometheus-node-exporter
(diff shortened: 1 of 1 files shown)


andy-paine:
This works for now but I think we should consider a couple of things going forward:
1. Should node exporter have an 'allow by default' model on the filesystem (my concern is that we will end up adding more to this sed replacement as we find more stuff it _shouldn't_ read)
2. Ancillary services like node exporter and filebeat - should we run them on the VM itself or should those also be scheduled by ECS?

These are slightly more existential questions than are relevant to this particular fix though 👍

PR #30, opened 2019-01-16
Pull Prometheus config from S3 and make Prometheus use it if it has changed

issyl0:
Edit: This is better now. Will deploy and test tomorrow.

https://trello.com/c/L8yEpoMc/114-verify-hub-dont-rebuild-prometheus-instances-on-config-change

PR #36, opened 2019-01-17
Prometheus config tweaks

philandstuff:
## use separate job label for node_exporter jobs

This appends `_node` to node_exporter jobs, so that they are given
names like `policy_node`.  This makes it easier to see when all of a
single type of metric exporter is broken.

## only keep the instances we're expecting
The static-ingress instances confused the `apps` scrape_config and
added targets for the static-ingress which would never successfully be
scraped.

If this scrape_config is just for apps, we should use `action: keep`
to only scrape app servers.


tlwr:
lets merge and deploy this after #30

issyl0:
That'll be a good test. 😬

PR #38, opened 2019-01-17
BAU: Use "drop-in" directory instead of using sed

adityapahuja:
It is not a good to use sed to change the default file (/etc/default/prometheus-node-exporter) provided by prometheus-node-exporter. Instead, this commits creates a "drop-in" directory prometheus-node-exporter.service.d which will be used to alter configuration settings for a unit without having to modify unit files.

Author: @adityapahuja

### DIFF
--- a/terraform/modules/hub/modules/ecs_asg/files/cloud-init.sh
+++ b/terraform/modules/hub/modules/ecs_asg/files/cloud-init.sh
@@ -152,6 +152,16 @@ docker run \
   amazon/amazon-ecs-agent:v1.23.0
 
 apt-get install --yes prometheus-node-exporter
-sed -i 's/sys|proc|dev|run/sys|proc|dev|run|var\/lib\/docker/' /etc/default/prometheus-node-exporter
+mkdir /etc/systemd/system/prometheus-node-exporter.service.d
+# Create an environment file for prometheus node exporter
+cat >  /etc/systemd/system/prometheus-node-exporter.service.d/prometheus-node-exporter.env <<EOF
+ARGS="--collector.diskstats.ignored-devices=^(ram|loop|fd|(h|s|v|xv)d[a-z]|nvme\\d+n\\d+p)\\d+$ --collector.filesystem.ignored-mount-points=^/(sys|proc|dev|run|var/lib/docker)($|/) --collector.netdev.ignored-devices=^lo$ --collector.textfile.directory=/var/lib/prometheus/node-exporter"
+EOF
+# Create an override file which will override prometheus node exporter service file
+cat > /etc/systemd/system/prometheus-node-exporter.service.d/10-override-args.conf <<EOF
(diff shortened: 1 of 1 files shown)


terraform/modules/hub/modules/ecs_asg/files/cloud-init.sh
```diff
@@ -152,6 +152,16 @@ docker run \
   amazon/amazon-ecs-agent:v1.23.0
 
 apt-get install --yes prometheus-node-exporter
-sed -i 's/sys|proc|dev|run/sys|proc|dev|run|var\/lib\/docker/' /etc/default/prometheus-node-exporter
+mkdir /etc/systemd/system/prometheus-node-exporter.service.d
+# Create an environment file for prometheus node exporter
+cat >  /etc/systemd/system/prometheus-node-exporter.service.d/prometheus-node-exporter <<EOF
```
tlwr:
```suggestion
cat >  /etc/systemd/system/prometheus-node-exporter.service.d/prometheus-node-exporter.conf <<EOF
```

tlwr:
```suggestion
cat >  /etc/systemd/system/prometheus-node-exporter.service.d/prometheus-node-exporter.env <<EOF
```

terraform/modules/hub/modules/ecs_asg/files/cloud-init.sh
```diff
@@ -152,6 +152,16 @@ docker run \
...
+ARGS="--collector.diskstats.ignored-devices=^(ram|loop|fd|(h|s|v|xv)d[a-z]|nvme\\d+n\\d+p)\\d+$ --collector.filesystem.ignored-mount-points=^/(sys|proc|dev|run|var/lib/docker)($|/) --collector.netdev.ignored-devices=^lo$ --collector.textfile.directory=/var/lib/prometheus/node-exporter"
+EOF
+# Create an override file which will override prometheus node exporter service file
+cat > /etc/systemd/system/prometheus-node-exporter.service.d/10-override-args.conf <<EOF
+[Service]
+EnvironmentFile=/etc/systemd/system/prometheus-node-exporter.service.d/prometheus-node-exporter
```
tlwr:
```suggestion
EnvironmentFile=/etc/systemd/system/prometheus-node-exporter.service.d/prometheus-node-exporter.env
```

tlwr:
Personal preference is that env files end in `.env`

terraform/modules/hub/modules/ecs_asg/files/cloud-init.sh
adityapahuja: Done. Default environment file (`/etc/default/prometheus-node-exporter`) from `prometheus-node-exporter` package does not have the filename extension (.env).

PR #48, opened 2019-01-21
Add an S3 bucket for verify-frontend app assets

issyl0:
- This is for storage of assets generated by `verify-frontend` deployments, ie CSS/JS/images.
- It's private because we're planning on putting CloudFront in front of it and have that serve the objects.
- Objects stay in this bucket for 365 days (a year) and then are "expired".

https://trello.com/c/uEJY5iZj/101-verify-hub-build-assets-and-ship-to-cloudfront-inside-pipeline

### DIFF
--- a/terraform/modules/hub/s3.tf
+++ b/terraform/modules/hub/s3.tf
@@ -5,6 +5,13 @@ resource "random_string" "deployment_config_bucket_name_suffix" {
   number  = false
 }
 
+resource "random_string" "verify_frontend_assets_bucket_name_suffix" {
+  length  = 6
+  special = false
+  upper   = false
+  number  = false
+}
+
 resource "aws_s3_bucket" "deployment_config" {
   bucket = "gds-${var.deployment}-config-${random_string.deployment_config_bucket_name_suffix.result}"
 
@@ -21,4 +28,9 @@ resource "aws_s3_bucket" "deployment_config" {
       }
     }
   }
+}
+
+resource "aws_s3_bucket" "verify_frontend_assets" {
+  bucket = "gds-${var.deployment}-verify-frontend-app-assets-${random_string.verify_frontend_assets_bucket_name_suffix.result}"
+  acl    = "private"
 }
--- a/terraform/modules/hub/s3.tf
+++ b/terraform/modules/hub/s3.tf
@@ -43,3 +43,23 @@ resource "aws_s3_bucket" "verify_frontend_assets" {
     }
   }
 }
+
+resource "aws_s3_bucket_policy" "verify_frontend_assets" {
+  bucket = "${aws_s3_bucket.verify_frontend_assets.id}"
+
+  policy = <<-POLICY
+  {
+    "Version": "2012-10-17",
+    "Statement": [
+      {
+        "Action": [
+          "s3:GetObject"
+        ]
+        "Effect": "Allow",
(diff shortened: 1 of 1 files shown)


terraform/modules/hub/s3.tf
```diff
@@ -22,3 +29,17 @@ resource "aws_s3_bucket" "deployment_config" {
     }
   }
 }
+
+resource "aws_s3_bucket" "verify_frontend_assets" {
+  bucket = "gds-${var.deployment}-verify-frontend-app-assets-${random_string.verify_frontend_assets_bucket_name_suffix.result}"
+  acl    = "private"
```
sakisv: If we want to restrict people going directly to the bucket we would probably need to follow something like this: https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/PrivateContent.html

However, since we're talking about CSS/js/etc we could probably go simpler with a `public-read` policy and also give that policy on the objects: 
https://docs.aws.amazon.com/AmazonS3/latest/dev/example-bucket-policies.html#example-bucket-policies-use-case-2

We should also configure CORS headers on the bucket: https://docs.aws.amazon.com/AmazonS3/latest/dev/cors.html

tlwr:
```suggestion
  acl    = "public-read"
```

issyl0: If we set the _whole bucket_ to `public-read`, do we need a `s3:GetObject` policy as well? That seems counter-intuitive to me?

issyl0: Done the public read policy on the objects in  e0f1b46.

Do we have prior art for CORS anywhere?

tlwr: I'm just thinking right now that this might be better done in `verify-terraform` in `environments/tools`; we aren't going to build and ship the assets into multiple environments. They can all reference the same bucket...

issyl0: Aren't we? What if we want to test asset changes in eg joint but run a different version in production?

tlwr: yeah but I was thinking our assets would be accessible on the path:
 `s3://name-of-bucket/${verify_frontend_commit_hash}-${verify_frontend_federation_config}/`

then we don't have to do promotion but it doesn't go to all environments at once, then the version number is passed in

tlwr: i.e. we have a `latest` and a `$commit_hash` bucket

issyl0: my thinking was that we'd do some asset bundling stuff in the Concourse job and push the objects to S3 on deploy. So they'd be `s3://name-of-bucket/asset1-randomrailshash.png" or somesuch (I've not yet thought about this extensively), so when we deploy joint we'd have those assets, then the same process would happen again when we deploy staging, etc. etc. So every environment would have a separate bucket (like we do with Prometheus configs, for instance).

issyl0: @tlwr I think it would be good to pair on this. I've a) never touched `verify-terraform`, b) it's a Verify repo and c) I don't want to get wires crossed, so it would be good to get this out together in the intended way.

issyl0:
This is superseded by https://github.com/alphagov/verify-terraform/pull/767.

PR #57, opened 2019-01-25
Remove the envvars with hardcoded paths to assets

issyl0:
- The app now gets its assets directly from an S3 bucket.

### DIFF
--- a/terraform/modules/hub/files/tasks/frontend.json
+++ b/terraform/modules/hub/files/tasks/frontend.json
@@ -66,6 +66,9 @@
       "Name": "COUNTRY_DISPLAY_LOCALES",
       "Value": "/verify-frontend/federation/configuration/display-locales/countries"
     }, {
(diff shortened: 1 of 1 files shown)


terraform/modules/hub/files/tasks/frontend.json
```diff
@@ -65,21 +65,6 @@
     }, {
       "Name": "COUNTRY_DISPLAY_LOCALES",
       "Value": "/verify-frontend/federation/configuration/display-locales/countries"
-    }, {
-      "Name": "COUNTRY_FLAGS_DIRECTORY",
-      "Value": "/verify-frontend/federation/configuration/eidas/country-flags"
-    }, {
-      "Name": "EIDAS_SCHEMES_DIRECTORY",
```
issyl0: Schemes is a different thing, not logos. Will restore...

PR #58, opened 2019-01-25
Add route53 record internally for policy redis

tlwr:
so our configuration can be the same between environments


tlwr:
i'm on the fence about doing this because it breaks TLS
we can make it an environment variable from SSM / parameter store, might be cleaner

PR #63, opened 2019-01-29
Add `reboot_required` as a metric that Prometheus collects

issyl0:
- We need to know when hosts require rebooting (for updates etc.).
- This adds a script that runs every minute to check for the presence of
  the reboot-required file and if it's there, update the Prometheus
  `node_reboot_required` metric to `1`. The textfile collector is
  already set up to read this directory.

Co-authored-by: @sakisv

### DIFF
--- a/terraform/modules/hub/modules/ecs_asg/files/cloud-init.sh
+++ b/terraform/modules/hub/modules/ecs_asg/files/cloud-init.sh
@@ -158,3 +158,22 @@ EOF
 systemctl daemon-reload
 systemctl enable prometheus-node-exporter
 systemctl restart prometheus-node-exporter
+
+cat <<EOF > /usr/bin/instance-reboot-required-metric.sh
+#!/usr/bin/env bash
+
+echo '# HELP node_reboot_required Node reboot is required for software updates.'
+echo '# TYPE node_reboot_required gauge'
+if [[ -f '/run/reboot-required' ]] ; then
+  echo 'node_reboot_required 1'
+else
+  echo 'node_reboot_required 0'
+fi
+EOF
+
+chmod +x /usr/bin/instance-reboot-required-metric.sh
+
+cat <<EOF | crontab -
+$(crontab -l | grep -v 'no crontab')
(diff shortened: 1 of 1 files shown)


terraform/modules/hub/modules/ecs_asg/files/cloud-init.sh
```diff
@@ -158,3 +158,24 @@ EOF
...
+else
+  echo 'node_reboot_required 0'
+fi
+EOF
+
+chmod +x /usr/bin/instance-reboot-required-metric.sh
+
+apt-get install --yes moreutils
+
+cat <<EOF | crontab -
```
philandstuff:
```suggestion
crontab - <<EOF
```
we don't need `cat` here 🚫 🐈

terraform/modules/hub/modules/ecs_asg/files/cloud-init.sh
```diff
@@ -158,3 +158,24 @@ EOF
...
+$(crontab -l | grep -v 'no crontab')
+* * * * * /usr/bin/instance-reboot-required-metric.sh | sponge /var/lib/prometheus/node-exporter/reboot-required.prom
```
philandstuff:
```suggestion
*/5 * * * * /usr/bin/instance-reboot-required-metric.sh | sponge /var/lib/prometheus/node-exporter/reboot-required.prom
```
do we need to run this every minute? the suggestion reduces it to every 5 minutes. I suspect we don't need to be super responsive to reboot-required checks so turning this down slightly won't have a negative impact on us.

terraform/modules/hub/modules/ecs_asg/files/cloud-init.sh
issyl0: But I like cats... 🐈

terraform/modules/hub/modules/ecs_asg/files/cloud-init.sh
sakisv: I would argue we could even run it once per hour because as far as I can tell it runs once a day at a random time

PR #81, opened 2019-02-06
Move security group config into separate files

tlwr:
What
----

We have security groups dotted around the place but also this random file
security_groups.tf...not anymore

How to review
-------------

Read the commits


issyl0:
Makes sense to put them in the projects to whcih they relate, I suppose.

PR #84, opened 2019-02-07
Correct predict-linear alert.

matthewcullum-gds:
There was a missing \ from the new filesystem exclusion query that hadn't been deployed. This now plans correctly and will apply.

Single: matthewcullum-gds


issyl0:
Does it still not apply with the square brackets?

matthewcullum-gds:
It applies either way - I have re-instated them

PR #94, opened 2019-02-11
AUTO-96: Parameterise app image tags

andy-paine:
- We want to be able to trace exactly what image went into a task
  definition so that we can reproduce issues etc. and for audit
- Create new variables and interpolate them into the taskdef
- Variable names are fairly descriptive so didn't put and description to
  keep the `variables.tf` file easier to read
- As these will have to be provided by the pipeline, they will come as
  variables to the `site.tf` so don't default anything here


tlwr:
see my comments on the verify-infra-config pr

PR #102, opened 2019-02-12
AUTO-172: Parameterize matomo site ID

andy-paine:
- We want to send all analytics data to a single Matomo instance so each
  environment needs a different site ID
- Named things Matomo as far as possible but need to use `PIWIK_SITE_ID`
  in the frontend config
- Defaulted so as to allow us to migrate each site individually

### DIFF
--- a/terraform/modules/hub/variables.tf
+++ b/terraform/modules/hub/variables.tf
@@ -53,6 +53,11 @@ variable "zendesk_username" {
   description = "Username (email address) for Zendesk access"
 }
 
+variable "matomo_site_id" {
+  description = "Site ID to use for Matomo"
+  default = 1
+}
+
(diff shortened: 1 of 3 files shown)


terraform/modules/hub/variables.tf
```diff
@@ -53,6 +53,11 @@ variable "zendesk_username" {
   description = "Username (email address) for Zendesk access"
 }
 
+variable "matomo_site_id" {
+  description = "Site ID to use for Matomo"
+  default = 1
```
sakisv: Should we have defaulted this to a non-prod value?

andy-paine: We probably should have, but we can also remove the default now `infrastructure-config` is merged 👍

PR #107, opened 2019-02-12
Fix prometheus

philandstuff:
Prometheus cloudinit was failing with
"/var/lib/prometheus/node-exporter/node-creation-time.prom: no such
file or directory".

I think these lines were missing

### DIFF
--- a/terraform/modules/hub/files/cloud-init/prometheus.sh
+++ b/terraform/modules/hub/files/cloud-init/prometheus.sh
@@ -92,6 +92,16 @@ systemctl restart journalbeat
 
 echo 'Installing prometheus node exporter'
 apt-get install --yes prometheus-node-exporter
+mkdir /etc/systemd/system/prometheus-node-exporter.service.d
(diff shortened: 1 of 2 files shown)


philandstuff:
Testing locally - I seem to have fixed node-exporter but ecs-agent didn't start. investigating

philandstuff:
i worked it out

it's because `/var/lib/prometheus/node-exporter` is created by the prometheus-node-exporter package but then we mount another filesystem at `/var/lib/prometheus` which hides the directory

philandstuff:
now prometheus itself won't start because it has different permissions to node-exporter >_<

matthewcullum-gds:
I think this could be cured by re-ordering the cloud-init and installing the node exporter after prometheus. I'll try this first thing and then raise a PR.

issyl0:
Can we have a better commit message than "wip" for 7cd5448?

tlwr:
I would prefer a better commit message too, Issy's suggestion is good

Alternatively use the squash and merge button

PR #115, opened 2019-02-13
Add dockerfile for running terraform

tlwr:
This dockerfile with the awscli and terraform

Again using the gdsre docker org until we have a better solution to open
source utility images


issyl0:
Can we make the Terraform version a variable so we only have to update it once instead of twice?

Although we’ll have to update the sha on every upgrade so I suppose it makes little difference.

tlwr:
Yes it is now a variable

PR #118, opened 2019-02-13
Segregate accessible cidr variable

tlwr:
Our prod environment should be world accessible but the mgmt lb shouldnt
be, we need separate variables for this


issyl0:
Why did you remove the default of "world accessible (0.0.0.0/0) for the existing variable?

tlwr:
It is a bad default IMO and I think was only there to ensure that the type was `list`

PR #125, opened 2019-02-14
Create new analytics alert.

matthewcullum-gds:
We need to raise an alert if the matomo analytics box is not responding.

Co-authored-by: @issyl0


issyl0:
Amended for commit message and changing hardcoded `staging` to `${deployment}` in the query.

issyl0:
Deployed. 🚀

PR #126, opened 2019-02-14
add $deployment as tag to dashboard

philandstuff:
so that we can display prod alerts separately from non-prod


philandstuff:
(deployed)

PR #130, opened 2019-02-15
AUTO-229: Update SSM to send logs to S3 bucket

adityapahuja:
This commit creates a new private S3 bucket with its native server side encryption enabled and its versioning is enabled. It has a bucket policy which allows instances to write session log files to the bucket and denies everyone from deleting the files from the bucket. It also updates AWS Systems Manager to start sending session log files to the bucket.

Co-authored-by: Athanasios Voutsadakis <athanasios.voutsadakis@digital.cabinet-office.gov.uk>

### DIFF
--- a/terraform/modules/account/ssm.tf
+++ b/terraform/modules/account/ssm.tf
@@ -1,0 +1,104 @@
+resource "aws_s3_bucket" "ssm_session_logs_store" {
+  bucket = "${var.deployment}-ssm-session-logs-store"
+  acl    = "private"
+  region = "eu-west-2"
+
+  versioning {
+    enabled = true
+  }
+
+  server_side_encryption_configuration {
+    rule {
+      apply_server_side_encryption_by_default {
+        sse_algorithm = "AES256"
+      }
+    }
+  }
+}
+
+data "aws_iam_policy_document" "ssm_session_logs_store" {
+  statement {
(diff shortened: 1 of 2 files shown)


terraform/modules/account/ssm.tf
```diff
@@ -0,0 +1,104 @@
+resource "aws_s3_bucket" "ssm_session_logs_store" {
+  bucket = "${var.deployment}-ssm-session-logs-store"
```
tlwr:
```suggestion
  bucket = "gds-${var.deployment}-ssm-session-logs-store"
```

tlwr:
The way this is implemented gives me some concerns which could be mitigated by putting this in a new module (which I hadn't foreseen, sorry)...

Basically if this is in the account module, and we provision a new account. Then we will have to deploy the account module twice, but will not be able to, without the root accounts.

This is because we will deploy the account module, there will be no roles provisioned yet - as there are no instances, then we deploy the hub module, and there are instances, but we have to get the root account out to deploy it again.

Conversely, because this looks at the instances to get the roles (via instance profiles), when we deprovision the old environments, the bucket policy will contain invalid principals, which may cause problems.

I'm not sure how valid these concerns are, or if there are mitigations. So I'd like some of your thoughts before we proceed further.

tlwr:
I'm actually fine with this

PR #135, opened 2019-02-18
Analytics proxy

sakisv:
Updates the INTERNAL_PIWIK_HOST for frontend task and makes the analytics endpoint a variable.

### DIFF
--- a/terraform/modules/hub/analytics.tf
+++ b/terraform/modules/hub/analytics.tf
@@ -12,7 +12,7 @@ resource "aws_security_group" "analytics_task" {
 locals {
   analytics_location_blocks = <<-LOCATIONS
 
-  set $analytics "https://analytics.tools.signin.service.gov.uk";
+  set $analytics "$ANALYTICS_ENDPOINT";
   location /analytics {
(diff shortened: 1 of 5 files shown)


terraform/modules/hub/analytics.tf
```diff
@@ -12,7 +12,7 @@ resource "aws_security_group" "analytics_task" {
 locals {
   analytics_location_blocks = <<-LOCATIONS
 
-  set $analytics "https://analytics.tools.signin.service.gov.uk";
+  set $analytics "$ANALYTICS_ENDPOINT";
```
tlwr:
```suggestion
  set $analytics "${analytics_endpoint}";
```

terraform/modules/hub/files/tasks/analytics.json
```diff
@@ -15,6 +15,10 @@
       {
         "Name": "LOCATION_BLOCKS",
         "Value": "${location_blocks_base64}"
+      },
```
tlwr: this is not needed, you can template it directly as above

PR #147, opened 2019-02-27
Add SAML reliability metrics to dashboard

tlwr:
Commit
------

```
This shows the % of requests at the SAML {,SOAP} Proxy endpoint which
are 2xx vs everything else.

There is probably a more prom way to do this

Signed-off-by: Toby Lorne <toby.lornewelch-richards@digital.cabinet-office.gov.uk>
```

What
----

Adds SAML endpoint reliability metrics to dashboard.

Why
---

So we can if we are getting failures on our SAML endpoints. This is a good indicator that there should be an error in sentry.

Preview
-------

screen shot 2019-02-27 at 09 09 18

Here it is on the dashboard, colocated with the general 5xx responses panel
screen shot 2019-02-27 at 09 11 25

Comment
-------

This might be unneeded because as you can see from above it is showing similar information to the 5xx and 2xx panels which it is next to. However the 3 endpoints I have chosen are good diagnostics for hub journeys.

### DIFF
--- a/dashboards/Hub-ECS.json
+++ b/dashboards/Hub-ECS.json
@@ -16,7 +16,7 @@
   "gnetId": null,
   "graphTooltip": 1,
   "id": 5,
-  "iteration": 1551083830301,
+  "iteration": 1551257520804,
   "links": [],
   "panels": [
     {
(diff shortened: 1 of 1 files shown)


dashboards/Hub-ECS.json
```diff
@@ -1339,6 +1339,193 @@
...
+      "pointradius": 5,
+      "points": false,
+      "renderer": "flot",
+      "seriesOverrides": [],
+      "spaceLength": 10,
+      "stack": false,
+      "steppedLine": false,
+      "targets": [
+        {
+          "expr": "sum(uk_gov_ida_hub_samlproxy_resources_SamlMessageReceiverApi_handleRequestPost_2xx_responses_total) / (sum(uk_gov_ida_hub_samlproxy_resources_SamlMessageReceiverApi_handleRequestPost_1xx_responses_total + uk_gov_ida_hub_samlproxy_resources_SamlMessageReceiverApi_handleRequestPost_2xx_responses_total + uk_gov_ida_hub_samlproxy_resources_SamlMessageReceiverApi_handleRequestPost_3xx_responses_total +\nuk_gov_ida_hub_samlproxy_resources_SamlMessageReceiverApi_handleRequestPost_4xx_responses_total +\nuk_gov_ida_hub_samlproxy_resources_SamlMessageReceiverApi_handleRequestPost_5xx_responses_total))",
```
philandstuff: you need a `rate` function on your counters here https://www.robustperception.io/rate-then-sum-never-sum-then-rate

You're computing y/x but you want dy/dx

tlwr: I also had a separate conversation with Andy about how this is pointless for SSP
and if we have the 5xxs for everything (except frontend 🙈 ) then this PR is pointless

PR #149, opened 2019-02-27
set prometheus web.external-url flag

philandstuff:
This sets the prometheus --web.external-url flag, so that when we
raise alerts to alertmanager (and on to pagerduty and slack), the
alerts have metadata and links that link back to the original
prometheus that raised them.

This is troublesome because we need to specify the external URL
differently for different prometheus instances, which means they need
different container definitions, which means they need different task
definitions, which means we need separate ECS services for each
prometheus.

To make as few changes as possible, I still run the services as DAEMON
services, but I put placement constraints on the task definitions so
that each task (and hence each service) is pinned to a specific AZ.

### DIFF
--- a/terraform/modules/hub/files/tasks/prometheus.json
+++ b/terraform/modules/hub/files/tasks/prometheus.json
@@ -30,7 +30,7 @@
     "entryPoint": [
       "sh",
       "-c",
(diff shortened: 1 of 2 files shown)


philandstuff:
🤢

PR #150, opened 2019-02-27
add all verify P1 alerts

philandstuff:
This adds all the verify P1 alerts (ie all alerts that actually go to
P1 pagerduty) from our grafana page-type alerts dashboard.

They need to be labelled with `severity: "p1"` in order to be routed
to pagerduty in alertmanager (although alertmanager is configured to
only send `deployment: "prod"` alerts to pagerduty, not other
environments).

### DIFF
--- a/terraform/modules/hub/files/prometheus/alerts.yml
+++ b/terraform/modules/hub/files/prometheus/alerts.yml
@@ -15,6 +15,8 @@ groups:
 - name: service
   rules:
   - alert: HubSamlProxyErrorsReceivingRequest
+    labels:
+      severity: "p1"
     annotations:
       message: |
         It looks like users are having trouble starting sessions at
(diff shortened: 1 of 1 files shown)


philandstuff:
(I forgot to explain in the commit that `deployment: "prod"` comes from prometheus's `external_labels`, and shouldn't be configured in the alerting rules here.)

terraform/modules/hub/files/prometheus/alerts.yml
```diff
@@ -15,17 +15,68 @@ groups:
...
       message: |
         It looks like users are having trouble starting sessions at
         the hub.  We expect that the saml-proxy handleRequestPost
         endpoint should return a 2xx response under normal conditions.
         We are observing the rate of 2xx responses has dropped below
         95%.
+    expr: |
+      sum without(instance)(
+          rate(uk_gov_ida_hub_samlproxy_resources_SamlMessageReceiverApi_handleRequestPost_2xx_responses_total[1m]))
+      / sum without(instance)(]
```
issyl0: Looks like there’s a stray `]` here.

tlwr:
```suggestion
      / sum without(instance)(
```

PR #152, opened 2019-03-01
AUTO-240: Fix analytics nginx

adityapahuja:
This commit updates analytics nginx to forward http_x_forwarded_for to matomo so that matomo can use it to identify visitor's location and to filter out unwanted matomo requests.
This commit adds an optional field called log_format which allows nginx to specify log format. If log_format is not defined, nginx config file will have the following comment '#log_format is not defined' and will use nginx's default log format.

Author: @adityapahuja

### DIFF
--- a/dockerfiles/nginx-tls/docker-entrypoint.sh
+++ b/dockerfiles/nginx-tls/docker-entrypoint.sh
@@ -5,6 +5,10 @@ location_blocks="${LOCATION_BLOCKS:?LOCATION_BLOCKS not set}"
 location_blocks="$(echo "$location_blocks" | base64 -d)"
 export location_blocks
 
+log_format="${LOG_FORMAT:-I2xvZ19mb3JtYXQgaXMgbm90IGRlZmluZWQK}"
+log_format="$(echo "$log_format" | base64 -d)"
+export log_format
+
 resolver="${RESOLVER:-10.0.0.2}"
 export resolver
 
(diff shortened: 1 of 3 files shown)


dockerfiles/nginx-tls/docker-entrypoint.sh
```diff
@@ -5,6 +5,10 @@ location_blocks="${LOCATION_BLOCKS:?LOCATION_BLOCKS not set}"
 location_blocks="$(echo "$location_blocks" | base64 -d)"
 export location_blocks
 
+log_format="${LOG_FORMAT:-I2xvZ19mb3JtYXQgaXMgbm90IGRlZmluZWQK}"
+log_format="$(echo "log_format" | base64 -d)"
```
philandstuff:
```suggestion
log_format="$(echo "$log_format" | base64 -d)"
```

philandstuff:
```suggestion
log_format="$(base64 -d <<<"$log_format")"
```

adityapahuja: Fixed. 👍

philandstuff: pick either or, not both :)

adityapahuja: I would like keep the change consistent since line 5 is already using echo command and it is easier to understand.

PR #155, opened 2019-03-04
HAProxy TLS termination

tlwr:
Context
-------

We need to present the user's source IP to frontend and analytics in order for the billing and audit system, as well as the analytics system.

HAProxy currently proxies the TCP connection it receives from the NLB to the ALB (in order for us to have managed static IP at ingress). However this means the source IP is the IP of HAproxy and not the user.

What
----

- New Docker image static-ingress-tls which is HAProxy but does TLS and HTTP (we still use the other one for forwarding to the ALB for 301s).

- Use the new static-ingress-tls image for HTTPS ingress

- Configure the NLB to treat static-ingress-https target group as a TLS upstream (rather than TCP)

- Configure the NLB to terminate TLS on port 443

How to review
------------

- Code review

### DIFF
--- a/terraform/modules/hub/static_ingress.tf
+++ b/terraform/modules/hub/static_ingress.tf
@@ -129,7 +129,7 @@ module "static_ingress_ecs_roles" {
   deployment       = "${var.deployment}"
   tools_account_id = "${var.tools_account_id}"
   service_name     = "static-ingress"
-  image_name       = "verify-static-ingress"
+  image_name       = "verify-static-ingress*"
 }
 
 resource "aws_ecs_task_definition" "static_ingress_http" {
--- a/terraform/modules/hub/static_ingress.tf
+++ b/terraform/modules/hub/static_ingress.tf
@@ -115,7 +115,7 @@ data "template_file" "static_ingress_https_task_def" {
   template = "${file("${path.module}/files/tasks/static-ingress.json")}"
 
   vars {
-    image_identifier = "${local.tools_account_ecr_url_prefix}-verify-static-ingress@${var.static_ingress_image_digest}"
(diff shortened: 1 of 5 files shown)


philandstuff:
Can you explain: if the NLB terminates TLS, how does it forward the client IP?

tlwr:
Yes, so the NLB's ip address is the users IP address, this is `magic`

These docs: https://docs.aws.amazon.com/elasticloadbalancing/latest/network/load-balancer-target-groups.html

> Source IP Preservation
> If you specify targets using an instance ID, the source IP addresses of the clients are preserved and provided to your applications.

Our static-ingress-https tasks have target group target type `instance` so the source IP is the user's IP

tlwr:
I thought this was in the commit but is in the card:

https://aws.amazon.com/blogs/aws/new-tls-termination-for-network-load-balancers/

terraform/modules/hub/static_ingress.tf
```diff
@@ -129,7 +129,7 @@ module "static_ingress_ecs_roles" {
   deployment       = "${var.deployment}"
   tools_account_id = "${var.tools_account_id}"
   service_name     = "static-ingress"
-  image_name       = "verify-static-ingress"
+  image_name       = "verify-static-ingress*"
```
sakisv: A comment to clarify why a `*` here is ok:

```suggestion
  # This is used in an IAM Policy document, so wildcards are ok
  image_name       = "verify-static-ingress*"
```

adityapahuja: Do we need a star in the image_name?

tlwr: yes, resolving this due to sakis's comment

PR #163, opened 2019-03-07
Decrease resolution of saml reliability: 1m -> 15m

tlwr:
1m is too noisy, 15m is what we had before


issyl0:
This == #162.

tlwr:
mornings

PR #164, opened 2019-03-07
AUTO: Add redirect to GOV.UK when accessing /

andy-paine:
- We used to do this redirect as per the service manual
- The service manual has since updated to suggest that this is not
  needed anymore but we will have that discussion later
- Restore old behaviour

Co-authored-by: Matthew Cullum <matthew.cullum@digital.cabinet-office.gov.uk>

### DIFF
--- a/terraform/modules/hub/ingress.tf
+++ b/terraform/modules/hub/ingress.tf
@@ -177,6 +177,27 @@ resource "aws_lb_listener" "ingress_https" {
   }
 }
 
+resource "aws_lb_listener_rule" "ingress_root" {
+  listener_arn = "${aws_lb_listener.ingress_https.arn}"
+  priority     = 140
+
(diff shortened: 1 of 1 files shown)


terraform/modules/hub/ingress.tf
```diff
@@ -177,6 +177,27 @@ resource "aws_lb_listener" "ingress_https" {
   }
 }
 
+resource "aws_lb_listener_rule" "ingress_root" {
+  listener_arn = "${aws_lb_listener.ingress_https.arn}"
+  priority     = 140
```
philandstuff: thank you for not picking 130 and colliding with #161 😁

andy-paine: seemed only right given I stole the rest of it off you

PR #168, opened 2019-03-07
AUTO: Only alert if saml stays unreliable

andy-paine:
- We were previously testing "has the success % of SAML requests dropped
  below 95% in the last 15 minutes"
- We want to test "has the success % of SAML requests stayed below 95%
  for the last 15 minutes"
- This is basically testing if the max is below 95% for 15 mins so just
  swap min -> max

### DIFF
--- a/terraform/modules/dashboards/pages_dashboard.json.tpl
+++ b/terraform/modules/dashboards/pages_dashboard.json.tpl
@@ -40,7 +40,7 @@
             },
             "reducer": {
               "params": [],
-              "type": "min"
+              "type": "max"
             },
             "type": "query"
           },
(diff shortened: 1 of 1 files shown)


issyl0:
Ahh, we've not moved this out of Grafana yet, that's why I couldn't find it when I was looking for alertmanager?

issyl0:
Is this still relevant given Toby has disabled Grafana alerting?

andy-paine:
https://prom-1.mgmt.signin.service.gov.uk/alerts suggests that they have been moved?

issyl0:
I'm confused.

issyl0:
Closing this as these alerts have been removed from Grafana in #172.

PR #173, opened 2019-03-09
Fire if threshold is breached for 15m

sakisv:
This alert was firing as soon as the threshold was breached and
because overnight the traffic is low each non-200 response had a
greater impact percentage-wise.

This commit adds a `for` statement[1] where prometheus will wait 15m before
firing the alert.

1: https://prometheus.io/docs/prometheus/latest/configuration/alerting_rules/#defining-alerting-rules

### DIFF
--- a/terraform/modules/hub/files/prometheus/alerts.yml
+++ b/terraform/modules/hub/files/prometheus/alerts.yml
@@ -30,6 +30,7 @@ groups:
       / sum without(instance)(
           rate(uk_gov_ida_hub_samlproxy_resources_SamlMessageReceiverApi_handleRequestPost_count[15m]))
       < 0.95
+    for: 15m
   - alert: HubSamlProxyErrorsReceivingResponse
     labels:
       severity: "p1"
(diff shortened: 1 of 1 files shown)


philandstuff:
:thinking: so first of all this alert is definitely too noisy and we need to fix it. 

That said, I’m generally not sure that `for: 15m` mixes well with `rate(metric[15m])` because it combines two different ways of computing alerts over a window - I find it easier to think about if we use one or the other. (And hence if we want to make it less noisy we increase the `rate` window: say, to 30m, rather than adding a `for`)

andy-paine:
That may be true but that can be discussed and dealt with in hours - I think this mostly brings this alert back to the old behaviour and should stop it waking people up in the middle of the night which I think is the scope of this PR.

philandstuff:
I thought the service was in maintenance mode overnight to stop the waking-people-up problem? (Per slack)

philandstuff:
Superseded by #176

PR #177, opened 2019-03-13
Reject DEBUG request method on frontend

sakisv:
This restores the functionality introduced here:
https://github.com/alphagov/verify-puppet/pull/1460

Co-authored-by: issyl0 <isabell.long@digital.cabinet-office.gov.uk>

### DIFF
--- a/terraform/modules/hub/hub_frontend.tf
+++ b/terraform/modules/hub/hub_frontend.tf
@@ -36,6 +36,11 @@ resource "aws_security_group_rule" "frontend_task_egress_to_internet_over_https"
 
 locals {
   location_blocks = <<-LOCATIONS
+  # Reject DEBUG verb in all cases.
+  if ($request_method = DEBUG ) {
+      return 405;
+  }
+
   location / {
     proxy_pass http://localhost:8080;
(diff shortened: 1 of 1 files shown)


terraform/modules/hub/hub_frontend.tf
```diff
@@ -36,6 +36,11 @@ resource "aws_security_group_rule" "frontend_task_egress_to_internet_over_https"
 
 locals {
   location_blocks = <<-LOCATIONS
+  # Reject DEBUG verb in all cases.
+  if ($request_method = DEBUG ) {
```
issyl0: Yes, the single `=` for comparison [is correct](http://nginx.org/en/docs/http/ngx_http_rewrite_module.html#if).

issyl0: There's an extra whitespace after `DEBUG` though.

issyl0:
```suggestion
  if ($request_method = DEBUG) {
```

issyl0:
Context for this change is in https://github.com/alphagov/verify-puppet/pull/1460#discussion_r232598986.

PR #189, opened 2019-03-20
Go back to old behaviour for SAML proxy metric alerts

tlwr:
Use 1m window and `for: 15m` in hub saml alerts

We want to be alerted if in the last 15 minutes if our reliability is <
95%

Andy and I went through the old behaviour with phil and phil applied
prometheus brain to the problem. There is probably a better
behaviour/config which requires discussion with verify

### DIFF
--- a/terraform/modules/hub/files/prometheus/alerts.yml
+++ b/terraform/modules/hub/files/prometheus/alerts.yml
@@ -14,31 +14,71 @@ groups:
       severity: "constant"
 - name: service
   rules:
-  - alert: HubSamlProxyNonEidasErrorRate
+  - alert: HubSamlProxyErrorsReceivingRequest
     labels:
       severity: "p1"
     annotations:
       message: |
-        It looks like users are having trouble with SAML interactions
-        in their journeys at the hub.  This means that they will be
-        unable to complete their journeys.  We expect that the
-        saml-proxy endpoints should return a 2xx response under normal
-        conditions.  We are observing the rate of 2xx responses has
-        dropped below 95%.
+        It looks like users are having trouble starting sessions at
+        the hub.  We expect that the saml-proxy handleRequestPost
+        endpoint should return a 2xx response under normal conditions.
+        We are observing the rate of 2xx responses has dropped below
+        95%.
     expr: |
       sum without(instance)(
-            rate(uk_gov_ida_hub_samlproxy_resources_SamlMessageReceiverApi_handleRequestPost_2xx_responses_total[15m])
-          + rate(uk_gov_ida_hub_samlproxy_resources_SamlMessageReceiverApi_handleResponsePost_2xx_responses_total[15m])
-          + rate(uk_gov_ida_hub_samlproxy_resources_SamlMessageSenderApi_sendJsonAuthnRequestFromHub_2xx_responses_total[15m])
-          + rate(uk_gov_ida_hub_samlproxy_resources_SamlMessageSenderApi_sendJsonAuthnResponseFromHub_2xx_responses_total[15m]))
+          rate(uk_gov_ida_hub_samlproxy_resources_SamlMessageReceiverApi_handleRequestPost_2xx_responses_total[15m]))
       / sum without(instance)(
-            rate(uk_gov_ida_hub_samlproxy_resources_SamlMessageReceiverApi_handleRequestPost_count[15m])
-          + rate(uk_gov_ida_hub_samlproxy_resources_SamlMessageReceiverApi_handleResponsePost_count[15m])
-          + rate(uk_gov_ida_hub_samlproxy_resources_SamlMessageSenderApi_sendJsonAuthnRequestFromHub_count[15m])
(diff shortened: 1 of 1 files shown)


philandstuff:
This is the old config which was noisy and woke people when there was no action to take.

The problem is that if you have fewer than 20 requests in a 15 minute period on a single endpoint, then a single 500 is enough to trigger the alert and wake someone.

tlwr:
This PR still doesn't implement the old behaviour which we think was correct, and the 1 5xx over 20 requests comment which Phil raised is correct. However this configuration will not have alerted anyone over the time period for which we have available data.

[Here](https://prom-3.mgmt.signin.service.gov.uk/graph?g0.range_input=2h&g0.end_input=2019-03-09%2001%3A09&g0.expr=sum%20without(instance)%20(rate(uk_gov_ida_hub_samlproxy_resources_SamlMessageReceiverApi_handleResponsePost_2xx_responses_total%5B15m%5D))%20%2F%20sum%20without(instance)%20(rate(uk_gov_ida_hub_samlproxy_resources_SamlMessageReceiverApi_handleResponsePost_count%5B15m%5D))&g0.tab=0&g1.range_input=2h&g1.end_input=2019-03-09%2001%3A09&g1.expr=sum(sum_over_time(uk_gov_ida_hub_samlproxy_resources_SamlMessageReceiverApi_handleResponsePost_2xx_responses_total%5B15m%5D))%20without%20(instance)%20%2F%20sum(sum_over_time(uk_gov_ida_hub_samlproxy_resources_SamlMessageReceiverApi_handleResponsePost_count%5B15m%5D))%20without%20(instance)%20&g1.tab=0) is an example of the old alert where it alerts, vs the updated configuration where it does not.

However what would be optimal would be getting back to the original behaviour agreed with Verify tech leadership (after an incident) of m1 rate being below 0.95 for 15 mins

Which in this case would be

```
expr: sum(sum_over_time(uk_gov_ida_hub_samlproxy_resources_SamlMessageReceiverApi_handleResponsePost_2xx_responses_total[1m])) without (instance) / sum(sum_over_time(uk_gov_ida_hub_samlproxy_resources_SamlMessageReceiverApi_handleResponsePost_count[1m])) without (instance) 
for: 15m
```

So [here](https://prom-3.mgmt.signin.service.gov.uk/graph?g0.range_input=2h&g0.end_input=2019-03-09%2001%3A09&g0.expr=sum%20without(instance)%20(rate(uk_gov_ida_hub_samlproxy_resources_SamlMessageReceiverApi_handleResponsePost_2xx_responses_total%5B15m%5D))%20%2F%20sum%20without(instance)%20(rate(uk_gov_ida_hub_samlproxy_resources_SamlMessageReceiverApi_handleResponsePost_count%5B15m%5D))&g0.tab=0&g1.range_input=2h&g1.end_input=2019-03-09%2001%3A09&g1.expr=sum(sum_over_time(uk_gov_ida_hub_samlproxy_resources_SamlMessageReceiverApi_handleResponsePost_2xx_responses_total%5B15m%5D))%20without%20(instance)%0A%2F%0Asum(sum_over_time(uk_gov_ida_hub_samlproxy_resources_SamlMessageReceiverApi_handleResponsePost_count%5B15m%5D))%20without%20(instance)%20&g1.tab=0&g2.range_input=2h&g2.end_input=2019-03-09%2001%3A09&g2.expr=sum(sum_over_time(uk_gov_ida_hub_samlproxy_resources_SamlMessageReceiverApi_handleResponsePost_2xx_responses_total%5B1m%5D))%20without%20(instance)%0A%2F%0Asum(sum_over_time(uk_gov_ida_hub_samlproxy_resources_SamlMessageReceiverApi_handleResponsePost_count%5B1m%5D))%20without%20(instance)%20&g2.tab=0) is a graph which shows all three options:

1. Noisy behaviour which we don't like
1. PR behaviour which is naive and wakes us up if 1 5xx over < 20 requests
1. Prometheus implementation of the old behaviour of m1 rate 

I am going to push a new commit to this PR which uses the third behaviour.

tlwr:
Phil and I have gone through this and think, for now, that this is correct behaviour.

I've updated the description.

terraform/modules/hub/files/prometheus/alerts.yml
```diff
@@ -14,31 +14,75 @@ groups:
...
-        It looks like users are having trouble with SAML interactions
-        in their journeys at the hub.  This means that they will be
-        unable to complete their journeys.  We expect that the
-        saml-proxy endpoints should return a 2xx response under normal
-        conditions.  We are observing the rate of 2xx responses has
-        dropped below 95%.
+        It looks like users are having trouble starting sessions at
+        the hub.  We expect that the saml-proxy handleRequestPost
+        endpoint should return a 2xx response under normal conditions.
+        We are observing the rate of 2xx responses has dropped below
```
galund: My point is slightly pedantic, and I assume irrelevant for what saml-proxy does, but anyway: we should count 3xx as successful too? Really only 4xx and 5xx should count as errors?

tlwr: SAML Proxy does not return `3xx`s

PR #191, opened 2019-03-25
He3ErfOp: create Verify frontend dashboard

bjgill:
View the Verify COncourse pipeline and also the error rate

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,1 +1,13 @@
 # verify-infrastructure
+
+## Create a new dashboard
+
+To create a new Grafana dashboard:
+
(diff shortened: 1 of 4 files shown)


bjgill:
OK, thanks. I'll do this and see I can find somewhere to document this for future reference.

bjgill:
OK - I've updated and added docs. Which of this and the verify-terraform PR need to go in first?

philandstuff:
> OK - I've updated and added docs. Which of this and the verify-terraform PR need to go in first?

PR #200, opened 2019-03-28
HAproxy improvements

tlwr:
- Use the [persist](https://www.haproxy.org/download/1.9/doc/configuration.txt) option
- JSON log format


tlwr:
After delving through the docs a bit more we instead changed the configuration to have a single backend, and use `option persist`

philandstuff:
I don't think `option persist` is enough on its own - i think we need multiple server blocks (as I think the PR originally had before the force push)?

PR #202, opened 2019-03-28
AUTO: Reduce max healthy %age for static ingress

andy-paine:
- These tasks bind to a host port so it doesn't make any sense to be
  able to scale beyond 1 per host
- By limiting the maximum to 100% and the minimum to 50% this will mean
  that ECS will remove an old task before trying to place a new one
- The minimum healthy percent means that there will always be 1 (or 2 in
  prod-like envs) tasks available so this is fine

This has been causing our "wait for ecs" task to timeout waiting for these to be placed which is causing unnecessary build failures


tlwr:
also makes instancing rolling faster :tm:

PR #206, opened 2019-03-29
Fix typo - remove a bad backslash from an EOF marker

issyl0:


issyl0:
Something something needle in a haystack... thanks @tlwr.

PR #207, opened 2019-04-05
AUTO: Alert on instance age instead of reboots

andy-paine:
- We run the ECS agent using docker run in cloud-init rather than
  putting it into systemd
- If we were to reboot a box, it would not bring the ECS agent back up
  so we shouldn't alert anyone telling them to reboot
- To get the unattended upgrades, we can just wait for a new AMI since
  we roll the instances every day anyway
- Instead of alerting on the reboot required file, let's use the
  node_creation_time metric we already have to detect the last time an
instance was rolled as this is an automated process
- This probably doubles up on alerting since the jobs on Concourse fail
  if the instances don't roll but they aren't cronitored so this catches
the case when they never run at all

### DIFF
--- a/terraform/modules/hub/files/cloud-init/prometheus.sh
+++ b/terraform/modules/hub/files/cloud-init/prometheus.sh
@@ -5,7 +5,6 @@ export DEBIAN_FRONTEND=noninteractive
 
 # Apt
 apt-get update  --yes
-apt-get upgrade --yes
 
 # AWS SSM Agent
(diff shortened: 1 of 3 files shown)


philandstuff:
We had a discussion IRL about this.  I'm a little concerned about removing the RebootRequired check, because a) I don't really understand what the release cadence of AMIs is (relative to security.ubuntu.com updates), and b) we don't have any monitoring that checks that we're on the latest AMI.

PR #212, opened 2019-05-02
EID-1422 Add splunk appender config to saml engine

Wynndow:
[There's a PR](https://github.com/alphagov/verify-hub/pull/268) to add a
Splunk appender to saml-engine. This appender needs config. This is that
config. [See this PR](https://github.com/alphagov/verify-infrastructure-config/pull/126) for the config that's stored in
verify-infrastructure-config.


Wynndow:
I did indeed. Why a split second? Doesn't the terraform need planning/applying or does it happen on merge?

andy-paine:
This module is now incompatible with the `site.tf` files in infra-config. If you merge infra-config first then it will be providing undeclared variables so it is a bit of a catch-22. If you merge them close together, you can probably beat the concourse resource check timer. Worst case scenario you get a red build.

PR #221, opened 2019-06-05
EID-1519 Model v1 A/B test

JohnWattsGDS:
Add RULES_DIRECTORY_VARIANT for AB test.

Added variable for ab_test_file - verify-infrastructure-config will need merging before this.

### DIFF
--- a/terraform/modules/hub/files/tasks/frontend.json
+++ b/terraform/modules/hub/files/tasks/frontend.json
@@ -88,6 +88,9 @@
       "Name": "RULES_DIRECTORY",
       "Value": "/verify-frontend/federation/configuration/idp-rules"
     }, {
+      "Name": "RULES_DIRECTORY_VARIANT",
+      "Value": "/verify-frontend/federation/configuration/idp-rules-variant"
+    }, {
       "Name": "RP_CONFIG",
       "Value": "/verify-frontend/federation/configuration/relying_parties.yml"
     }, {
@@ -98,7 +101,7 @@
       "Value": "/verify-frontend/federation/configuration/cycle-three-attributes"
     }, {
       "Name": "AB_TEST_FILE",
-      "Value": "/verify-frontend/federation/configuration/ab_test/ab_test.yml"
+      "Value": "/verify-frontend/federation/configuration/ab_test/${ab_test_file}"
     }, {
       "Name": "SEGMENT_DEFINITIONS",
       "Value": "/verify-frontend/federation/configuration/segment_definitions.yml"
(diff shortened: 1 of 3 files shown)


terraform/modules/hub/files/tasks/frontend.json
```diff
@@ -99,6 +102,9 @@
     }, {
       "Name": "AB_TEST_FILE",
       "Value": "/verify-frontend/federation/configuration/ab_test/ab_test.yml"
```
jakubmiarka: you'll want to do something like this
```suggestion
      "Value": "/verify-frontend/federation/configuration/ab_test/${ab_test_file}"
```
and then in add it here https://github.com/alphagov/verify-infrastructure-config/blob/65b5c10bc4e39436b89e182ea7f5eb512bc18c59/terraform/deployments/staging/hub/site.tf#L48

However I'd suggest doing as two separate PRs - when tearing down the AB test it will be easier to revert if the PR has only the variant stuff ;)

terraform/modules/hub/files/tasks/frontend.json
```diff
@@ -99,6 +102,9 @@
...
+    }, {
+      "Name": "AB_TEST_FILE",
```
jakubmiarka: no need to duplicate it, see above.

JohnWattsGDS: I deleted that before committing and pushing.  I just forgot to add the changed file before committing and pushing 😄

terraform/modules/hub/files/tasks/frontend.json
JohnWattsGDS: Thanks for that @jakubmiarka !

Wynndow:
Did you see that your other PR related to this was reverted?

JohnWattsGDS:
Yeah, I've just seen @Wynndow.  Looks like I needed to do something in this repo too ... ?!?

philandstuff:
From a brief look, I think this PR probably needs to be merged first, then the verify-infrastructure-config one.

philandstuff:
This repo only contains terraform modules (ie shared library code); verify-infrastructure-config makes calls into this one via its `module` declarations.  So I suspect that alphagov/verify-infrastructure-config#146 broke because it was using a feature from this PR, but this PR hadn't been merged yet.

philandstuff:
🤔 looking at it a bit more closely, the two PRs need to be merged simulatneously to avoid breaking the build, because this introduces a new variable without giving it a default value.  It is probably better to use a default value to allow the changes without needing coordination; but to get this out today it's probably fine to just communicate in slack to let people know you are merging both and to cancel any builds that include one PR but not the other.

PR #222, opened 2019-06-10
Reboot at instance creation time in order to pick up new kernels

philandstuff:
Canonical do not bake new AMIs as frequently as they release kernel
patches.  Therefore, when we launch an instance, it may not have the
latest kernel.  This causes alert noise because we see lots of
RebootRequired alerts.

We need to ensure that everything we want to run will survive the
reboot.  We run most things as systemd services, so that will be
fine.  The ECS agent is run via docker.  If we set `--restart=always`,
then the docker daemon will restart the ECS agent on daemon start --
even after a reboot.

#How to review

- Check to see that I haven't missed any important services
- Merge at the start of a day and destroy a few instances in staging to ensure they pick up the new cloud init

### DIFF
--- a/terraform/modules/hub/modules/ecs_asg/files/cloud-init.sh
+++ b/terraform/modules/hub/modules/ecs_asg/files/cloud-init.sh
@@ -148,7 +148,7 @@ docker run \
   --privileged \
   --name ecs-agent \
   --detach=true \
-  --restart=on-failure:10 \
+  --restart=always \
   --volume=/etc/ecs:/etc/ecs \
   --volume=/lib64:/lib64 \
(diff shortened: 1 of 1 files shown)


szd55gds (approved):
Looks sensible to me

sakisv (approved):

sakisv:
In case anyone else has the same question as me:

I was wondering how would the `reboot` at the end of the script would help, given that cloud-init.sh only runs when the instance is created. 

@philandstuff pointed out that this will work because right at the beginning of the cloud-init script we do

```
apt-get update --yes
apt-get upgrade --yes
```

So, the new kernels will be installed there and reboot will apply them.

PR #224, opened 2019-06-11
Don't reboot on startup for now

philandstuff:
We have a problem where journalbeat doesn't start properly after a
reboot, so we lose logs :sadpanda:

This just stops the reboot for now while we investigate the issue.


philandstuff:
Superseded by #225

PR #226, opened 2019-06-12
AUTO: Reboot Prometheus on startup

andy-paine:
- This also makes the ECS agent restart always so that when the box
  reboots, the agent comes up again afterwards

Co-authored-by: Aditya Pahuja <aditya.pahuja@digital.cabinet-office.gov.uk>


philandstuff:
Do we also need to fix journalbeat (ie port #225 to prometheus)?

PR #229, opened 2019-06-13
prometheus cloud-init: Remount prometheus disk after reboot using label

Krenair:
Rather than device names, which do not persist across reboots.

We considered using UUID, however:
* This is neater
* This is going to be static anyway
* The label is just nicer for humans than the UUID

### DIFF
--- a/terraform/modules/hub/files/cloud-init/prometheus.sh
+++ b/terraform/modules/hub/files/cloud-init/prometheus.sh
@@ -152,7 +152,8 @@ if [ -z "$(lsblk | grep "$vol" | awk '{print $7}')" ] ; then
     echo "volume /dev/$vol is mounted ; mounting"
 
   if grep -qv "/dev/$vol" /etc/fstab ; then
-    echo "/dev/$vol /srv/prometheus ext4 defaults,nofail 0 2" >> /etc/fstab
+    e2label /dev/$vol srv-prometheus
+    echo "LABEL=srv-prometheus /srv/prometheus ext4 defaults,nofail 0 2" >> /etc/fstab
   fi
 fi
 


adityapahuja (approved):

terraform/modules/hub/files/cloud-init/prometheus.sh
```diff
@@ -152,7 +152,8 @@ if [ -z "$(lsblk | grep "$vol" | awk '{print $7}')" ] ; then
     echo "volume /dev/$vol is mounted ; mounting"
 
   if grep -qv "/dev/$vol" /etc/fstab ; then
-    echo "/dev/$vol /srv/prometheus ext4 defaults,nofail 0 2" >> /etc/fstab
+    e2label /dev/$vol srv-prometheus
```
sakisv: Does this depend on having `/dev/nvme1n1` be the "right" volume the first time that we boot?

Would it perhaps be safer if we did some bash magic and find the name of the device based on its size?

Krenair: I assume it's always the right volume on first boot, it will presumably have formatted the first one (nvme0n1) as the root device.

philandstuff: Certainly so far we haven't seen the problems on first boot; only on reboot

philandstuff (approved):

terraform/modules/hub/files/cloud-init/prometheus.sh
philandstuff: That said, it'd be good to verify that Prometheus gets the big disk and / gets the little one

sakisv (approved):

PR #230, opened 2019-06-13
Revert "prometheus cloud-init: Remount prometheus disk after reboot using label"

philandstuff:
Reverts alphagov/verify-infrastructure#229


philandstuff:
This cloud init failed and prometheus is dead ⚰️

philandstuff:
For context: the error was at the `e2label` line, and it said something along the lines of `Superblock checksum does not match superblock`

PR #234, opened 2019-06-18
Add config metadata bucket to Hub Staging for self service

Rosa-Fox:
- The self-service app will publish metadata to this s3 bucket, which hub config will then read from

Co-authored-by: Brendan Butler <brendan.butler@digital.cabinet-office.gov.uk>

### DIFF
--- a/terraform/modules/self-service/variables.tf
+++ b/terraform/modules/self-service/variables.tf
@@ -1,0 +1,3 @@
+variable "deployment" {
+  description = "Name of the deployment; {staging,prod,integration}"
+}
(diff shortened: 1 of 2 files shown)


terraform/modules/self-service/variables.tf
```diff
@@ -0,0 +1,3 @@
+variable "deployment" {
+  description = "Name of the deployment; {joint,staging,prod,integration}"
```
jakubmiarka: joint doesn't exist anymore
```suggestion
  description = "Name of the deployment; {staging,prod,integration}"
```

Rosa-Fox: Thanks @jakubmiarka have amended!

PR #235, opened 2019-06-18
prometheus cloud-init: Remount prometheus disk after reboot using UUIDs

Krenair:


terraform/modules/hub/files/cloud-init/prometheus.sh
```diff
@@ -149,10 +149,12 @@ if [ -z "$(lsblk | grep "$vol" | awk '{print $7}')" ] ; then
     echo "volume /dev/$vol is not mounted ; mounting"
     mount "/dev/$vol" /srv/prometheus
   fi
-    echo "volume /dev/$vol is mounted ; mounting"
+
+  echo "volume /dev/$vol is mounted ; writing fstab entry"
 
   if grep -qv "/dev/$vol" /etc/fstab ; then
-    echo "/dev/$vol /srv/prometheus ext4 defaults,nofail 0 2" >> /etc/fstab
+    UUID=$(ls -l /dev/disk/by-uuid | grep nvme1n1 | cut -d' ' -f9)
```
philandstuff:
```suggestion
    UUID=$(ls -l /dev/disk/by-uuid | grep $vol | cut -d' ' -f9)
```

terraform/modules/hub/files/cloud-init/prometheus.sh
```diff
@@ -149,10 +149,12 @@ if [ -z "$(lsblk | grep "$vol" | awk '{print $7}')" ] ; then
...
     mount "/dev/$vol" /srv/prometheus
   fi
-    echo "volume /dev/$vol is mounted ; mounting"
+
+  echo "volume /dev/$vol is mounted ; writing fstab entry"
 
   if grep -qv "/dev/$vol" /etc/fstab ; then
-    echo "/dev/$vol /srv/prometheus ext4 defaults,nofail 0 2" >> /etc/fstab
+    UUID=$(ls -l /dev/disk/by-uuid | grep $vol | cut -d' ' -f9)
+    echo "UUID=$UUID /srv/prometheus ext4 defaults,nofail 0 2" >> /etc/fstab
```
adityapahuja: Can we include uid in /etc/fstab? E.g.
```echo "UUID=$UUID /srv/prometheus ext4 defaults,uid=nobody,nofail 0 2" >> /etc/fstab``` 
or
```echo "UUID=$UUID /srv/prometheus ext4 defaults,uid=65534,nofail 0 2" >> /etc/fstab ```

Krenair: Possibly but not in this commit. Right now the instances appear to mount this as nobody:root on first boot and still mount it as nobody:root on reboots thereafter.

Krenair: Also to do that you would probably want to change the mount options on first boot above too.

adityapahuja: Please see line 162. The command changes the ownership of `/srv/prometheus` directory to nobody.

Krenair: Ah yes. In that case I assume there is no point specifying uid when mounting/in fstab?

PR #238, opened 2019-06-20
Add Self Service config metadata bucket to Hub VPC endpoint

brenetic:
The Self Service app will be uploading metadata to an S3 bucket in the hub's vpc and therefore needs to be added as a resource.

https://trello.com/c/3VTpnx6M/466-create-user-and-roles-for-s3-access-in-staging

### DIFF
--- a/terraform/modules/self-service/self_service_app.tf
+++ b/terraform/modules/self-service/self_service_app.tf
@@ -1,9 +1,14 @@
 data "aws_iam_policy_document" "self_service_user_write_to_bucket" {
   statement {
-    sid       = "AllowPutObject"
+    sid       = "AllowGetAndPutObject"
     effect    = "Allow"
     resources = ["${aws_s3_bucket.config_metadata.arn}/*"]
-    actions   = ["s3:PutObject"]
+    actions   = [
+      "s3:GetO*",
+      "s3:PutO*",
+      "s3:DeleteO*",
+      "s3:ListBucket",
+    ]
   }
 }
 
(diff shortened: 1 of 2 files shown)


terraform/modules/self-service/self_service_app.tf
```diff
@@ -20,3 +20,9 @@ resource "aws_iam_user_policy_attachment" "self_service_user_policy_attachment"
   user       = "${aws_iam_user.self_service_user.name}"
   policy_arn = "${aws_iam_policy.self_service_user_write_to_bucket.arn}"
 }
+
+resource "aws_vpc_endpoint" "self_service_s3" {
```
philandstuff: there is already an S3 vpc endpoint in `terraform/modules/hub/vpc.tf` - I'm not sure we need another one?

philandstuff: In particular, if we deploy the self-service app to the `internal` subnets, we'll get to reuse the S3 endpoint for free.

brenetic: The endpoint in hub has permissions to read as well as write. I guess I'm a little unsure whether limiting the connecting user to have read permissions only is enough?

The Self Service app won't be deployed to the Hub for a little while yet and in the mean time we would like to begin integrating the publishing of metadata from the app and hub config reading it.

philandstuff:
This is a good start. I'm guessing you'll put in another PR to give Config the permissions to pull from the new bucket?

brenetic:
Yes indeed. PR is here: https://github.com/alphagov/verify-infrastructure/pull/240

PR #247, opened 2019-07-03
FIREBREAK: Added the new verify-sentry dockerfile

j-maynard:
In order to use GitHub as a SSO provider for logging in to Sentry we need to install an additional plugin using `pip`.
This docker image is going to get build by our pipeline and put into ECR from which it will get deployed.

### DIFF
--- a/dockerfiles/verify-sentry/Dockerfile
+++ b/dockerfiles/verify-sentry/Dockerfile
@@ -1,0 +1,1 @@
+FROM sentry:9.0-onbuild
--- a/dockerfiles/verify-sentry/requirements.txt
+++ b/dockerfiles/verify-sentry/requirements.txt
@@ -1,0 +1,1 @@
+https://github.com/getsentry/sentry-auth-github/archive/43f6b270b3fac32326518a78be77562ebe5abacf.zip


dockerfiles/verify-sentry/Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM sentry:9.1-onbuild
```
jakubmiarka: we don't need to use the `on-build` variant if we're actually not making any use of it...

Also, I'd probably stay on 9.0 instead of upgrading to 9.1 to make the change smaller.

j-maynard: Switched to 9.0.

j-maynard:
[Use Github OAuth for Sentry](https://trello.com/c/gaQhv2Lh/243-use-github-oauth-for-sentry)

dockerfiles/verify-sentry/Dockerfile
```diff
@@ -0,0 +1,2 @@
+FROM sentry:9.0
+RUN pip install https://github.com/getsentry/sentry-auth-github/archive/master.zip
```
philandstuff: is it worth considering pinning a specific version rather than `master`?

jakubmiarka: the last commit was `Sep 22, 2017` though... might be worth just in case...

chrisholmes: Also, would a `requirements.txt` file be applicable here? I think GitHub should scan it to check for vulnerable dependencies.

philandstuff: 👍 that would also have the benefit of pinning all transitive dependencies, not just the sentry-auth-github one

jakubmiarka: sentry-auth-github doesn't have any dependencies... but yeah, we can use the requirements.txt file. And then we can switch to the -onbuild image...

j-maynard: Resolved this to use the requirements.txt and the onbuild version of Sentry.  Pegged the version of the Sentry-Auth-Github to the current commit version on Github.

PR #253, opened 2019-07-26
Self Service Fargatization in Staging

brenetic:
This adds all the infrastructure in order for the Self Service app.
Currently this will only be live in staging.

We are using the Verify Hub remote state in order to gain access to some
shared config such as private subnets and the Hub VPC id.

We also refactored some of the files a little bit to make them better
organised.

The app's `RAILS_ENV` is indeed set to `development` deliberately.

https://trello.com/c/QsDuRhZK/496-define-ecs-fargate-task-in-staging-deploying-to-staging-vpc

Screen Shot 2019-07-26 at 16 01 37

Co-authored-by: Jakub Miarka <jakub.miarka@digital.cabinet-office.gov.uk>


issyl0:
I never want to see the word "Fargatization" again.

PR #267, opened 2019-08-09
Allow Self Service task access to config metadata bucket

brenetic:
### Remove Self Service IAM user

This was used by the instance of the self service app that was deployed to the PaaS. It's no longer needed as the app is living in Hub Staging.

### Give Self Service Fargate task config metadata access

The entire purpose for the Self Service app to exist is to talk to the config metadata bucket. I guess we should allow this.

### Diff

Screen Shot 2019-08-09 at 10 54 48

https://trello.com/c/vJ5YZedB/589-add-missing-s3-permissions-to-the-self-service-task-role

### DIFF
--- a/terraform/modules/self-service/iam.tf
+++ b/terraform/modules/self-service/iam.tf
@@ -1,32 +1,3 @@
-data "aws_iam_policy_document" "self_service_user_write_to_bucket" {
-  statement {
-    sid       = "AllowGetAndPutObject"
-    effect    = "Allow"
-    resources = ["${aws_s3_bucket.config_metadata.arn}/*"]
-
-    actions = [
-      "s3:GetO*",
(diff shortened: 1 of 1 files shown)


terraform/modules/self-service/iam.tf
```diff
@@ -213,3 +184,28 @@ resource "aws_iam_role_policy_attachment" "execution_execution" {
   role       = "${aws_iam_role.self_service_execution.name}"
   policy_arn = "${aws_iam_policy.execution.arn}"
 }
+
+data "aws_iam_policy_document" "access_config_meta_data" {
```
jakubmiarka:
```suggestion
data "aws_iam_policy_document" "access_config_metadata" {
```
to be consistent?

PR #270, opened 2019-08-12
Add Sentry DSN secret to Self Service app

brenetic:
### What

We will be using Sentry with the Self Service app. We are happy for it to be bundled with the other hub apps.

The usual Sentry Raven gem change will need adding to the Self Service app itself as well.

https://trello.com/c/lKcAMBvI/528-plug-the-app-to-sentry

### DIFF
--- a/terraform/modules/self-service/ecs.tf
+++ b/terraform/modules/self-service/ecs.tf
@@ -11,6 +11,7 @@ locals  {
     cognito_user_pool_id  = "${aws_cognito_user_pool.user_pool.id}"
     asset_host            = "${var.asset_host}"
     asset_prefix          = "${element(split(":", var.image_digest),1)}/assets/"
+    sentry_dsn            = "arn:aws:ssm:eu-west-2:${data.aws_caller_identity.account.account_id}:parameter/${var.deployment}/${local.service}/sentry-dsn"
   }
 }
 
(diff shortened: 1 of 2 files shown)


terraform/modules/self-service/ecs.tf
```diff
@@ -11,6 +11,7 @@ locals  {
     cognito_user_pool_id  = "${aws_cognito_user_pool.user_pool.id}"
     asset_host            = "${var.asset_host}"
     asset_prefix          = "${element(split(":", var.image_digest),1)}/assets/"
+    sentry_dsn            = "arn:aws:ssm:eu-west-2:${data.aws_caller_identity.account.account_id}:parameter/${var.deployment}/ecs-app-shared/sentry-dsn"
```
jakubmiarka: self-service should have its own DSN, otherwise this will go to the hub project but we should create our own sentry project.

brenetic: Ah ok, I thought we were sharing. Will look at setting one up

jakubmiarka: sorry, I should have been clearer, we're sharing the sentry instance, but projects are different, hub, frontend and we need to add self-service...

brenetic: Okey dokey, have created a self service sentry DSN and stuck it in AWS SSM, FTW

PR #278, opened 2019-08-28
Adjust config metadata access policy

brenetic:
- Be more explicit about permissions
- Stop using the parameter store. The bucket ARN is probably not a secret.
- Add integration bucket to the S3 endpoint policy
- Remove the route table ids output as we no longer need to create a self service specific s3 endpoint

https://trello.com/c/PMZorhCM/634-add-permission-for-app-in-staging-to-publish-to-integration-bucket

### DIFF
--- a/terraform/modules/hub/vpc.tf
+++ b/terraform/modules/hub/vpc.tf
@@ -37,7 +37,9 @@ resource "aws_vpc_endpoint" "s3" {
                 "arn:aws:s3:::gds-${var.deployment}-ssm-session-logs-store",
                 "arn:aws:s3:::gds-${var.deployment}-ssm-session-logs-store/*",
                 "arn:aws:s3:::govukverify-self-service-${var.deployment}-config-metadata",
-                "arn:aws:s3:::govukverify-self-service-${var.deployment}-config-metadata/*"
+                "arn:aws:s3:::govukverify-self-service-${var.deployment}-config-metadata/*",
+                "arn:aws:s3:::govukverify-self-service-integration-config-metadata",
+                "arn:aws:s3:::govukverify-self-service-integration-config-metadata/*"
             ]
         }
     ]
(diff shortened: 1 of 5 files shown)


terraform/modules/self-service/iam.tf
```diff
@@ -190,15 +190,17 @@ data "aws_iam_policy_document" "access_config_metadata" {
     sid       = "AllowGetAndPutObject"
     effect    = "Allow"
     resources = [
+      "${aws_s3_bucket.config_metadata.arn}",
       "${aws_s3_bucket.config_metadata.arn}/*",
-      "${data.aws_ssm_parameter.integration_metadata_bucket.value}/*"
+      "arn:aws:s3:::govukverify-${local.service}-integration-config-metadata",
```
jakubmiarka: this will give access to the bucket from all the environments... should we pass it in per environment somehow?

terraform/modules/hub/vpc.tf
```diff
@@ -37,7 +37,9 @@ resource "aws_vpc_endpoint" "s3" {
                 "arn:aws:s3:::gds-${var.deployment}-ssm-session-logs-store",
                 "arn:aws:s3:::gds-${var.deployment}-ssm-session-logs-store/*",
                 "arn:aws:s3:::govukverify-self-service-${var.deployment}-config-metadata",
-                "arn:aws:s3:::govukverify-self-service-${var.deployment}-config-metadata/*"
+                "arn:aws:s3:::govukverify-self-service-${var.deployment}-config-metadata/*",
+                "arn:aws:s3:::govukverify-self-service-integration-config-metadata",
```
jakubmiarka: this also whitelist the S3 integration bucket in all environments 🤔

brenetic: We will address this again when we come to move the S3 bucket into it's own module in the hub

terraform/modules/self-service/site.tf
```diff
@@ -1,5 +1,9 @@
 locals {
-  service = "self-service"
+  service              = "self-service"
+  config_metadata_arns = [
```
jakubmiarka:
```suggestion
  config_metadata_buckets_arns = [
```
?

PR #282, opened 2019-08-29
DO NOT MERGE: temporary_password_validity_days

jkunle:
based on open PR: https://github.com/terraform-providers/terraform-provider-aws/pull/8845

temporary_password_validity_days can now be specified in a password policy in an aws_cognito_user_pool.
One gotcha is that to use the new field in combination with admin_create_user_config fields
you will need to specify the deprecated admin_create_user_config.unused_account_validity_days to 0. Or else you will get an AWS API error that advises you to use the new password_policy.temporary_password_validity_days field.

### DIFF
--- a/terraform/modules/self-service/cognito.tf
+++ b/terraform/modules/self-service/cognito.tf
@@ -7,7 +7,8 @@ resource "aws_cognito_user_pool" "user_pool" {
 
   admin_create_user_config {
     allow_admin_create_user_only = true
-    unused_account_validity_days = 1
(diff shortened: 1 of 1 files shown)


issyl0:
This needs a rebase to fix conflicts, if indeed this is still necessary/being worked on. Looks like the AWS Terraform provider isn't in a place to be merged yet.

jakubmiarka:
We have commited a workaround fix. Once the the provider is updated we'll update our code in-place, so this PR is no longer needed.

PR #286, opened 2019-08-30
Temporary Password Validity Days

jkunle:
Alternate implementation to the the trello card
[Set Cognito temporary password expiry to 1 day](https://trello.com/c/A0wglRIe/625-set-cognito-temporary-password-expiry-to-1-day)

The default is 7 days. For security reasons this should be set to the
minimum possible value, which is 1 day.


terraform/modules/self-service/cognito.tf
```diff
@@ -70,9 +70,13 @@ resource "aws_cognito_user_pool" "user_pool" {
     prevent_destroy = true
   }
 
-  provisioner "local-exec" {
+   provisioner "local-exec" {
```
jakubmiarka:
```suggestion
  provisioner "local-exec" {
```

PR #289, opened 2019-09-02
TfEzFC4W implement cross domain google analytics tracking config variables

chrisclayson:
## What

We need to implement some Google Analytics (GA) tracking code on the Hub.

This isn't because we're going to use GA for Verify analytics.

It's because GOV.UK is now implementing GA tracking across various x.gov.uk subdomains.

GDS needs to be able to track user journeys across campaign pages, GOV.UK content and transactions on the service domain.

Verify is part of these journeys, hence the need to get the GA code up and running on the Hub. 

## How
Add new variables to hub module that will contain GA config values.
Add cross domain config values to frontend task environment variables

Co-authored-by: Alex Gee <alexander.gee@digital.cabinet-office.gov.uk>

### DIFF
--- a/terraform/modules/hub/variables.tf
+++ b/terraform/modules/hub/variables.tf
@@ -89,6 +89,16 @@ variable "self_service_enabled" {
   default     = "false"
 }
 
+variable "cross_gov_ga_account_id" {
+  description = "The Google Analytics account ID for GOV.UK cross domain analysis"
+  default     = ""
+}
+
+variable "cross_gov_ga_domain_names" {
+  description = "List of (space delimited) domains to automatically enable links for cross-domain analytics"
+  default     = ""
+}
+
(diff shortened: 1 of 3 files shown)


terraform/modules/hub/variables.tf
```diff
@@ -89,6 +89,16 @@ variable "self_service_enabled" {
   default     = "false"
 }
 
+variable "cross_gov_ga_tracker_id" {
+  description = "The Google Analytics tracker ID for GOV.UK cross domain analysis"
+  default     = ""
+}
+
+variable "cross_gov_ga_domain_names" {
+  description = "List of (space delimited) domains to automatically enable links for cross-domain analytics"
```
galund: Also mention that this is used for SAML post forms or other cases where the user is redirected back to the original service.

PR #298, opened 2019-09-20
Remove password complexity for Self Service app

brenetic:
[NCSC guidelines](https://www.ncsc.gov.uk/collection/passwords/updating-your-approach#PasswordGuidance:UpdatingYourApproach-Donotusecomplexityrequirements) recommend that we do not enforce complex passwords for users, only a minimum number of characters.

Screen Shot 2019-09-20 at 11 47 29

https://trello.com/c/vfOQoNM3/676-remove-password-complexity

### DIFF
--- a/terraform/modules/self-service/modules/cognito/cognito.tf
+++ b/terraform/modules/self-service/modules/cognito/cognito.tf
@@ -104,10 +104,6 @@ resource "aws_cognito_user_pool" "user_pool" {
 
   password_policy {
     minimum_length    = 8
-    require_uppercase = true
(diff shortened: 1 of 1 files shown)


jakubmiarka:
> 8 characters seems really short though if we're allowing pure alphabetical passwords.
> 
> I assume we don't want to loosen the restrictions we have on temporary passwords.

8 characters is industry standard (Google, Microsoft, Yahoo)
6 characters are more common though (Amazon, LinkedIn, Facebook, Twitter)

Yep, the temporary password should keep its stricter conditions...

PR #309, opened 2019-10-11
cfoy7qoB: Hide the healthcheck from the public internet

jakubmiarka:
Added a new rule on to the loadbalancer returning a fixed response (401)
when path is /healthcheck.

Co-Authored-By: Rosa Fox <rosa.fox@digital.cabinet-office.gov.uk>

### DIFF
--- a/terraform/modules/self-service/lb.tf
+++ b/terraform/modules/self-service/lb.tf
@@ -66,3 +66,22 @@ resource "aws_lb_listener" "cluster_https" {
     target_group_arn = "${aws_lb_target_group.task.arn}"
   }
 }
+
+resource "aws_lb_listener_rule" "disable_health_check_from_public" {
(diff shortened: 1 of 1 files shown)


terraform/modules/self-service/lb.tf
```diff
@@ -66,3 +66,22 @@ resource "aws_lb_listener" "cluster_https" {
     target_group_arn = "${aws_lb_target_group.task.arn}"
   }
 }
+
+resource "aws_lb_listener_rule" "disable_health_check_from_public" {
+  listener_arn = "${aws_lb_listener.cluster_https.arn}"
```
brenetic: don't need to interpolate any more

jakubmiarka: oh yeah, everything else is still interpolated, so will do a wider clean-up exercise later

PR #312, opened 2019-10-24
BAU: Add a dependency between nginx and frontend

adityapahuja:
AWS ECS introduced Enhanced Container Dependency Management on 7th March 2019. This commit update frontend task definition to include a dependency between nginx and frontend. This will ensure that nginx comes up after the frontend is up and running.

Author: @adityapahuja


tlwr:
The ECS agent will probably have to be updated before this can be merged

PR #315, opened 2019-10-24
BAU: Fix JSON format error

adityapahuja:
AWS threw the following error.
`ECS Task Definition container_definitions is invalid: Error decoding JSON: json: cannot unmarshal array into Go struct field ContainerDefinition.HealthCheck of type ecs.HealthCheck`

According to AWS documentation (https://docs.amazonaws.cn/en_us/AWSCloudFormation/latest/UserGuide/aws-properties-ecs-taskdefinition-healthcheck.html), Command should have a string array. This commit updates to change Command's type from string to string array.

Author: @adityapahuja

### DIFF
--- a/terraform/modules/hub/files/tasks/frontend.json
+++ b/terraform/modules/hub/files/tasks/frontend.json
@@ -147,7 +147,7 @@
     }],
     "healthCheck" : [
       {
-        "Command": ["curl -sfm10 localhost:8080/cookies"],
+        "Command": [ "CMD-SHELL", "curl -sfm10 http://localhost:8080/cookies || exit 1" ],
         "Interval": 10,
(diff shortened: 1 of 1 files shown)


terraform/modules/hub/files/tasks/frontend.json
```diff
@@ -147,7 +147,7 @@
     }],
     "healthCheck" : [
       {
-        "Command": "curl -sfm10 localhost:8080/cookies",
+        "Command": ["curl -sfm10 localhost:8080/cookies"],
```
Krenair:
```suggestion
        "Command": ["curl", "-sfm10", "localhost:8080/cookies"],
```

PR #319, opened 2019-11-01
Updated e-mail template

j-maynard:
Updated the e-mail template to include HTML tags for new lines.

See Trello: https://trello.com/c/pLO92N0S

### DIFF
--- a/terraform/modules/self-service/modules/cognito/cognito.tf
+++ b/terraform/modules/self-service/modules/cognito/cognito.tf
@@ -104,24 +104,24 @@ resource "aws_cognito_user_pool" "user_pool" {
     invite_message_template {
       email_subject = "You have been invited to collaborate on the GOV.UK Verify Manage certificates service"
       email_message = <<-EOT
-        Dear {username}<br>
-        <br>
-        You have been invited to collaborate on the GOV.UK Verify Manage certificates service.<br>
-        <br>
-        Sign in at ${var.domain} using the following temporary password:<br>
-        <br>
-        <strong>{####}</strong><br>
-        <br>
-        You will be asked to create a new password and set up multi-factor authentication using your preferred authentication app.<br>
-        <br>
-        <span style="color: #d4351c">Please sign in within 24 hours, otherwise the temporary password will expire.</span><br>
-        <br>
-        If you miss this deadline, contact your admin to ask for another temporary password.<br>
-        <br>
-        Thanks<br>
-        <br>
-        The GOV.UK Verify team<br>
-        <a href="https://www.verify.service.gov.uk/">https://www.verify.service.gov.uk/</a>
+        <p>Dear {username}</p>
+        
+        <p>You have been invited to collaborate on the GOV.UK Verify Manage certificates service.</p>
+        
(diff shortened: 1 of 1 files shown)


terraform/modules/self-service/modules/cognito/cognito.tf
```diff
@@ -104,24 +104,24 @@ resource "aws_cognito_user_pool" "user_pool" {
...
+        <br>
+        You have been invited to collaborate on the GOV.UK Verify Manage certificates service.<br>
+        <br>
+        Sign in at ${var.domain} using the following temporary password:<br>
+        <br>
+        <strong>{####}</strong><br>
+        <br>
+        You will be asked to create a new password and set up multi-factor authentication using your preferred authentication app.<br>
+        <br>
+        <span style="#d4351c">Please sign in within 24 hours, otherwise the temporary password will expire.</span><br>
```
galund: That doesn't look like CSS to me. Is it supposed to be?

j-maynard: No... Its straight HTML as its sent from AWS.  So there are limits to what to include in the e-mail template in terraform.

j-maynard: But there is a mistake there ;-)

galund: I can clarify what I meant: the contents of a `style` attribute must be CSS, and this isn't.

brenetic:
I think we should use `<p>` tags instead of line breaks. Line breaks go inside `<p>` tags if memory serves

terraform/modules/self-service/modules/cognito/cognito.tf
```diff
@@ -104,24 +104,24 @@ resource "aws_cognito_user_pool" "user_pool" {
...
+        
+        <p>You have been invited to collaborate on the GOV.UK Verify Manage certificates service.</p>
+        
+        <p>Sign in at ${var.domain} using the following temporary password:</p>
+        
+        <p style="font-weight: bold;">{####}</p>
+        
+        <p>You will be asked to create a new password and set up multi-factor authentication using your preferred authentication app.</p>
+        
+        <p style="color: #d4351c">Please sign in within 24 hours, otherwise the temporary password will expire.</p>
```
galund: I imagine that we should use `<strong>` with a style rather than just styling `<p>`, but it need block you merging.

terraform/modules/self-service/modules/cognito/cognito.tf
```diff
@@ -104,24 +104,24 @@ resource "aws_cognito_user_pool" "user_pool" {
...
-
-        If you miss this deadline, contact your admin to ask for another temporary password.
-
-        Thanks
-
-        The GOV.UK Verify team
-        https://www.verify.service.gov.uk/
+        <p>Dear {username}</p>
+        
+        <p>You have been invited to collaborate on the GOV.UK Verify Manage certificates service.</p>
```
galund: The capitalisation here is a bit wacky. If we are capitalising the name of the service, then the C and the S should be capitals too. If not, then the M should not be capitalised.

PR #320, opened 2019-11-04
BAU: Specify CPU and Memory requirements

adityapahuja:
According to Amazon Elastic Container Service documentation, 1 virtual CPU is equal to 1024 CPU shares. Setting CPU to 0 in a container definition means that it needs 2 CPU shares. This CPU requirement is quite low. This commit updates all container definitions and task definitions to specify both CPU and memory requirements. This will ensure that each application gets a fair share of CPU and memory.

Author: @adityapahuja


adityapahuja:
Another advantage is that specifying the CPU and memory requirements in task definition will enable us to easily migrate from EC2 to Fargate. Fargate requires CPU and memory to be specified in task definition. 

Source: https://docs.aws.amazon.com/AmazonECS/latest/developerguide/AWS_Fargate.html

PR #321, opened 2019-11-04
BAU: nginx changes

adityapahuja:
This pull request has 2 commits.

The first commit adds more information about upstream system in nginx log file. This will help us to find out the cause of 502 errors. It also removes upstream_cookie_name because it has no value and according to Nginx documentation upstream_cookie_name needs to be defined e.g. upstream_cookie_x_govuk_session_cookie.

The second commit updates nginx config file to match nginx config's settings used in old Verify infrastructure (e.g. ssl_session_cache and client_body_buffer_size). client_body_buffer_size is set to 64k because the default size for this is 16k on x64 and an IDP response can be approximately 27kb.

Author: @adityapahuja


dockerfiles/nginx-tls/docker-entrypoint.sh
```diff
@@ -14,13 +14,17 @@ log_format json_event '{ \"@timestamp\": \"\$time_iso8601\", '
...
+                         '\"upstream_response_time\": \"\$upstream_response_time\", '
                          '\"upstream_response_length\": \"\$upstream_response_length\", '
                          '\"upstream_bytes_received\": \"\$upstream_bytes_received\", '
+                         '\"upstream_bytes_sent\": \"\$upstream_bytes_sent\", '
+                         '\"upstream_connect_time\": \"\$upstream_connect_time\", '
+                         '\"upstream_queue_time\": \"\$upstream_queue_time\", '
+                         '\"upstream_session_time\": \"\$upstream_session_time\", '
+                         '\"upstream_header_time\": \"\$upstream_header_time\", '
                          '\"upstream_cache_status\": \"\$upstream_cache_status\", '
-                         '\"upstream_cookie_name\": \"\$upstream_cookie_name\", '
```
philandstuff: Did you intentionally remove this? There's nothing in your PR description or commit message which mentions removing fields so I just wanted to check if tis was a mistake?

dockerfiles/nginx-tls/nginx.conf.tpl
```diff
@@ -29,7 +29,10 @@ http {
     ssl_ciphers         ECDHE-RSA-AES128-GCM-SHA256:ECDHE-RSA-AES256-GCM-SHA384:ECDHE-RSA-AES128-SHA256:ECDHE-RSA-AES256-SHA384:ECDHE-RSA-AES256-GCM-SHA512:DHE-RSA-AES256-GCM-SHA512:DHE-RSA-AES256-GCM-SHA384;
     ssl_certificate     /tmp/tls/cert.pem;
     ssl_certificate_key /tmp/tls/key.pem;
-    ssl_session_cache   shared:SSL:1m;
+    ssl_session_cache   shared:SSL:2m;
+
+    client_max_body_size    5m;
```
philandstuff: the default is `1m`.  If the client sends a body bigger than `client_max_body_size`, nginx returns a `413` status code.  I strongly suspect that no normal client would ever need to exceed this but we can check if we've been returning many 413s.  If we don't need it I'm inclined to leave it at the default, because we don't want to allow malicious clients to send bigger requests than necessary.

dockerfiles/nginx-tls/nginx.conf.tpl
```diff
@@ -29,7 +29,10 @@ http {
...
+    client_body_buffer_size 50k;
```
philandstuff:
```suggestion
    client_body_buffer_size 64k;
```
I suggest we use a nice round number to line up with memory page boundaries.  [The default value is two pages](http://nginx.org/en/docs/http/ngx_http_core_module.html#client_body_buffer_size) which is 8k or 16k depending on the system.

dockerfiles/nginx-tls/docker-entrypoint.sh
adityapahuja: Yes, I added upstream_cookie_name but realised that it had no value because according to Nginx documentation upstream_cookie_`name` needs to be defined e.g.  upstream_cookie_`x_govuk_session_cookie`.

dockerfiles/nginx-tls/nginx.conf.tpl
adityapahuja: The nginx config used in old Verify infrastructure uses 2m

dockerfiles/nginx-tls/nginx.conf.tpl
adityapahuja: The nginx config used in old Verify infrastructure uses 50k for client_body_buffer_size. The intention of this pull request is to make nginx config used in new Verify infrastructure to have same settings as nginx config used in old Verify infrastructure.

dockerfiles/nginx-tls/nginx.conf.tpl
philandstuff: We spoke in person and we think we can leave it at 1m. Happy for this to be implicit or explicit.

dockerfiles/nginx-tls/nginx.conf.tpl
philandstuff: Discussed in person, we're happy with 64k.

adityapahuja: Done.

adityapahuja:
I have added explanations for these changes to the commit messages.

PR #338, opened 2019-12-02
HUH-234 add variables for 2019 short hub ABC test configuration

instantiator:
Adds:

* `RULES_B_DIRECTORY`
* `RULES_C_DIRECTORY`
* `ABC_VARIANTS_CONFIG`
* `SEGMENT_DEFINITIONS_VARIANT_B`
* `SEGMENT_DEFINITIONS_VARIANT_C`

### DIFF
--- a/terraform/modules/hub/files/tasks/frontend.json
+++ b/terraform/modules/hub/files/tasks/frontend.json
@@ -100,6 +100,12 @@
       "Name": "RULES_DIRECTORY_VARIANT",
       "Value": "/verify-frontend/federation/configuration/idp-rules-variant"
     }, {
+      "Name": "RULES_B_DIRECTORY",
+      "Value": "/verify-frontend/federation/configuration/idp-rules-variant-b"
+    }, {
+      "Name": "RULES_C_DIRECTORY",
+      "Value": "/verify-frontend/federation/configuration/idp-rules-variant-c"
+    }, {
       "Name": "RP_CONFIG",
       "Value": "/verify-frontend/federation/configuration/relying_parties.yml"
     }, {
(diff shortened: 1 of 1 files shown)


terraform/modules/hub/files/tasks/frontend.json
```diff
@@ -111,9 +117,18 @@
...
       "Value": "/verify-frontend/federation/configuration/ab_test/${ab_test_file}"
+    }, {
+      "Name": "ABC_VARIANTS_CONFIG",
+      "Value": "/verify-frontend/federation/configuration/special-cases/abc-variants.yml"
     }, {
       "Name": "SEGMENT_DEFINITIONS",
       "Value": "/verify-frontend/federation/configuration/segment_definitions.yml"
+    }, {
+      "Name": "SEGMENT_DEFINITIONS_VARIANT_B",
+      "Value": "/verify-frontend/federation/configuration/segment_definitions_variant_b.yml"
```
jakubmiarka: I guess variant B could use the control segment definitions, since it'll use the same segment, just not all of them...

instantiator: True - although in stub that wasn't true - and so there's code to support different configurations in each of the 3 tests (control A, variant B, variant C).

galund:
(Spoke to Brendan who says one could run the `plan` from a checkout of https://github.com/alphagov/verify-infrastructure-config )

jakubmiarka:
> Approving, but I'd like to have seen a `plan` before doing so, and there's nothing in the readme that suggests how to do that :-(

It's just a JSON file, terraform plan won't tell you much

PR #348, opened 2019-12-10
Adding temporary hub-environment-legacy variable

jkunle:
Adding hub-environment-legacy variable so that we can extend the
scheme of the hub-environment variable to accomodate more values
in a future PR.

Co-authored-by: Rosa Fox <Rosa.fox@digital.cabinet-office-gov.uk>

### DIFF
--- a/terraform/modules/self-service/variables.tf
+++ b/terraform/modules/self-service/variables.tf
@@ -69,6 +69,10 @@ variable "hub_host" {
   default     = "signin.service.gov.uk"
 }
 
+variable "hub_environments_legacy" {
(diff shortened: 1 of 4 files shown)


terraform/modules/self-service/variables.tf
```diff
@@ -69,6 +69,10 @@ variable "hub_host" {
   default     = "signin.service.gov.uk"
 }
 
+variable "hub_environments_legacy" {
+  description = "JSON string of hub environments and the config metadata buckets"
```
jakubmiarka: you could say in here it's temporary

brenetic:
Is this going to be removed soon?

PR #357, opened 2020-01-08
BAU: Resolve ECS task placement issue on ingress cluster

adityapahuja:
This pull request has two commits.

1. The first commit updates CPU and memory limits in all tasks for ingress cluster so that they can run on one EC2 instance. This will stop ECS from taking a long time to figure out which EC2 instance has enough CPU or memory for four tasks having different CPU and memory requirements.

2. The second commit removes placement constraints from frontend, analytics and metadata tasks for ingress cluster. ECS does not do task placement very well. It is better for all tasks to run on one EC2 instance so that ECS doesn't spend time figuring out which one of EC2 instances has enough CPU and memory space for each task (frontend, analytics, metadata and beat-exporter).

I ran terraform apply. There were no errors.

Author: @adityapahuja

### DIFF
--- a/terraform/modules/hub/analytics.tf
+++ b/terraform/modules/hub/analytics.tf
@@ -49,11 +49,6 @@ resource "aws_ecs_task_definition" "analytics" {
   container_definitions = data.template_file.analytics_task_def.rendered
   network_mode          = "awsvpc"
   execution_role_arn    = module.analytics_ecs_roles.execution_role_arn
-
-  placement_constraints {
-    type       = "memberOf"
-    expression = "not(task:group == service:${var.deployment}-frontend-v2)"
-  }
 }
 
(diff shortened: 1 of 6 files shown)


adityapahuja:
The spreasheet (https://docs.google.com/spreadsheets/d/1pr-eSNAYLQ7KJqR8uDx0ZWZs5qRiRL2430wGrm3PY3I/edit#gid=18868071) show how all 4 tasks fit on one EC2 instance.

adityapahuja:
Please note that one EC2 instance (set to t3.medium) has 3 network interfaces. Each task requires a network interface. This means that 4 tasks cannot fit on one EC2 instance. There are more EC2 instances (see https://github.com/alphagov/verify-infrastructure/commit/ef8d07be6dc9945778cf7c65bde592c80bb4f66f) . This should not be the issue.

PR #360, opened 2020-01-13
BAU gain more memory by changing from medium to large

idrop:
During self-assessment period, we are seeing a significant memory increase on `saml-engine`.
We can double our memory available by upgrading to t3.large (4GB to 8GB).


idrop:
we want to do this in prod only

philandstuff:
@idrop and on saml-engine only as well?

PR #370, opened 2020-01-31
Hub dashboard to show better indication of available memory

galund:
Buffers and caches can be freed if necessary, per Phil

### DIFF
--- a/dashboards/Hub-ECS.json
+++ b/dashboards/Hub-ECS.json
@@ -2115,7 +2115,7 @@
           ]
         },
         {
-          "alias": "EC2 Memory Free (GB)",
+          "alias": "EC2 Memory Available (GB)",
           "colorMode": "value",
           "colors": [
             "#890f02",
@@ -2170,7 +2170,7 @@
       ],
       "targets": [
         {
-          "expr": "node_memory_MemFree / 1024 / 1024 / 1024",
+          "expr": "node_memory_MemAvailable / 1024 / 1024 / 1024",
           "format": "table",
           "instant": true,
           "interval": "",


Wynndow:
Can you stick a screenshot of what the graph looks like with this change?

philandstuff:
@galund what would it take for this to stop being a draft?

galund:
@philandstuff dunno, if you're happy then I'll click the ready button (and apologies to @Wynndow because I forgot about his question, but I guess we're all happy now).

dashboards/Hub-ECS.json
```diff
@@ -2170,7 +2170,7 @@
       ],
       "targets": [
         {
-          "expr": "node_memory_MemFree / 1024 / 1024 / 1024",
+          "expr": "(node_memory_MemFree + node_memory_Buffers + node_memory_Cached) / 1024 / 1024 / 1024",
```
adityapahuja: How about node_memory_MemAvailable?

adityapahuja: For example, `node_memory_MemAvailable / 1024 / 1024 / 1024`

philandstuff: TIL. +1 to this suggestion, see also [the man page for free](http://manpages.ubuntu.com/manpages/xenial/en/man1/free.1.html) for more information, in particular:

             Estimation of how much memory is available for starting new  applications,  without
              swapping.  Unlike  the  data provided by the cache or free fields, this field takes
              into account page cache and also that not all  reclaimable  memory  slabs  will  be
              reclaimed  due  to  items being in use (MemAvailable in /proc/meminfo, available on
              kernels 3.14, emulated on kernels 2.6.27+, otherwise the same as free)

galund: 🉑

PR #372, opened 2020-02-03
Bump engine and rds cert

samcrang:
Merge #371 first (and rebase).

### DIFF
--- a/terraform/modules/self-service/db.tf
+++ b/terraform/modules/self-service/db.tf
@@ -1,7 +1,7 @@
 resource "aws_db_instance" "self_service" {
   name           = "selfservice"
   engine         = "postgres"
-  engine_version = "10.6"
+  engine_version = "11.5"
   storage_type   = "gp2"
 
   instance_class    = var.db_instance_class
@@ -32,7 +32,7 @@ resource "aws_db_instance" "self_service" {
   iam_database_authentication_enabled = true
   apply_immediately                   = true
 
-  ca_cert_identifier = "rds-ca-2015"
+  ca_cert_identifier = "rds-ca-2019"
 
   lifecycle {
     prevent_destroy = true


jakubmiarka:
The previous #371 is failing build now:
`Error: "ca_cert_identifier": this field cannot be set`
https://deployer.tools.signin.service.gov.uk/teams/main/pipelines/verify-self-service/jobs/deploy-to-staging/builds/487#L5e29bfb9:55

Do we need to update the provider first?

samcrang:
@jakubmiarka Yup—I've fixed that now so this should be good to go.

jakubmiarka:
> @jakubmiarka Yup—I've fixed that now so this should be good to go.

We need to go higher than the version you updated it too, because of this: https://github.com/terraform-providers/terraform-provider-aws/pull/10890

Until now, we've been using a fork: https://github.com/alphagov/verify-terraform-provider-aws

terraform/modules/self-service/db.tf
```diff
@@ -1,7 +1,7 @@
 resource "aws_db_instance" "self_service" {
   name           = "selfservice"
   engine         = "postgres"
-  engine_version = "10.6"
+  engine_version = "10.10"
```
jakubmiarka:
```suggestion
  engine_version = "11.5"
```
https://deployer.tools.signin.service.gov.uk/teams/main/pipelines/verify-self-service/jobs/deploy-to-production/builds/175#L5e29c042:146

samcrang: Hmmm. I wonder why production is running a major version newer version of Postgres compared to staging. I guess we can just do this change but it seem odd...

jakubmiarka: they were provisioned at different times (and the default version was different?).. would be my guess 🤷‍♂

PR #374, opened 2020-02-14
AUT-515: Resolve roll-staging failure

adityapahuja:
`roll-staging` job failed due to the following error messages.

E: Could not get lock /var/lib/dpkg/lock-frontend - open (11: Resource temporarily unavailable)
E: Unable to acquire the dpkg frontend lock (/var/lib/dpkg/lock-frontend), is another process using it?

This change updates the `cloud-init.sh` script to retry up to 5 times whenever it fails to acquire a lock. When the script fails to acquire the lock 5 times, it will exit with an error code.

Author: @adityapahuja


adityapahuja:
I have updated the script to identify one or more processes who owns the locks. This will make it easy for us to find the cause of the lock issues.

PR #380, opened 2020-03-23
BAU: Create a separate variable for prometheus

adityapahuja:
Prometheus needs to run in each Availability Zone (AZ). There are three Availability Zones in London region. There is no need to scale to more than 3 instances of Prometheus. This change creates a separate variable called number_of_prometheus_apps which defaults to 3. This replaces number_of_apps with number_of_prometheus_apps for Prometheus cluster only.

Author: @adityapahuja


adityapahuja:
Please do not merge this pull request until we terminate one bad instance of prometheus (due to cloudinit's mounting issue) in production.

PR #382, opened 2020-03-24
AUT-606: Update to support xlarge for frontend apps

adityapahuja:
This change

- adds a new task definition for Verify Frontend used in t3.xlarge instance type so that it can use more CPU and memory units.
- adds a conditional statement to ensure that it uses number of EC2 efficiently. This change is required because t3.xlarge supports 4 network interfaces whereas t3.medium supports 3 network interfaces. There are four different containers (analytics, frontend, metadata and beat-exporter). Each container needs a network interface to work properly. So t3.xlarge can support 4 containers whereas t3.medium supports 3 containers. That's why we do not need to double the number of EC2 instances for t3.xlarge.
- adds a conditional statement to ensure that it uses correct task definition for verify fontend for instance type. If the instance type is t3.xlarge, verify frontend app and its ngnix will have more CPU and memory units. If it is not t3.xlarge, verify frontend and its nginx will remain unchanged.
- replaces the generic variable `instance_type` with specific variable `ingress_instance_type`. This change allows us to choose instance type for applications running in ingress cluster only in different environment.

Other applications (e.g. metadata, analytics and beat-exporter) will remain unchanged. This change is needed for reducing the frontend application's CPU usage.

Co-authored-by: Phillip Miller <phillip.miller@digital.cabinet-office.gov.uk>

### DIFF
--- a/terraform/modules/hub/files/tasks/frontend_xlarge.json
+++ b/terraform/modules/hub/files/tasks/frontend_xlarge.json
@@ -1,0 +1,177 @@
+[
+  {
+    "name": "nginx",
+    "image": "${nginx_image_identifier}",
+    "cpu": 1024,
+    "memory": 1024,
+    "essential": true,
+    "portMappings": [
+      {
+        "containerPort": 8443,
+        "hostPort": 8443
+      }
+    ],
+    "environment": [
+      {
+        "Name": "LOCATION_BLOCKS",
+        "Value": "${location_blocks_base64}"
+      }
+    ],
+    "dependsOn": [
+      {
+        "containerName": "frontend",
+        "condition": "HEALTHY"
+      }
+    ]
+  },
+  {
+    "name": "frontend",
+    "image": "${image_identifier}",
+    "cpu": 2424,
+    "memory": 14066,
+    "essential": true,
+    "portMappings": [
+      {
+        "containerPort": 8080,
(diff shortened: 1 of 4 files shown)


terraform/modules/hub/files/tasks/frontend_xlarge.json
```diff
@@ -0,0 +1,171 @@
...
+      "Value": "true"
+    }, {
+      "Name": "CROSS_GOV_GOOGLE_ANALYTICS_TRACKER_ID",
+      "Value": "${cross_gov_ga_tracker_id}"
+    }, {
+      "Name": "CROSS_GOV_GOOGLE_ANALYTICS_DOMAIN_LIST",
+      "Value": "${cross_gov_ga_domain_names}"
+    }, {
+      "Name": "PUBLISH_HUB_CONFIG_ENABLED",
+      "Value": "${publish_hub_config_enabled}"
```
jakubmiarka: this has been updated on master, could you update it as per https://github.com/alphagov/verify-infrastructure/pull/381 ?

adityapahuja: This has been done.

adityapahuja: NOt yet. Let me update the new file.

PR #386, opened 2020-03-27
Alert about high CPU usage on ingress nodes

Krenair:


Krenair:
nb: we should probably be auto-scaling instead (or as well as?), this is a stopgap so we know the frontend is getting dangerously high load and we need to scale up

Krenair:
this isn't what we want

PR #387, opened 2020-03-30
Export ECS CPU utilisation for the frontend to Prometheus

Krenair:


terraform/modules/hub/files/prometheus/cloudwatch_exporter.yml
```diff
@@ -30,3 +30,12 @@ metrics:
     LoadBalancer:
       - 'app/.*-ingress.*'
   aws_statistics: [Sum]
+- aws_namespace: AWS/ECS
+  aws_metric_name: CPUUtilization
```
CharlesIC: If this is just for frontend, shouldn't the metric name have "frontend" in the name?

PR #388, opened 2020-03-30
Alert about high CPU usage by frontend service

Krenair:


terraform/modules/hub/files/prometheus/alerts.yml
```diff
@@ -166,5 +166,13 @@ groups:
     for: 20m
     expr: |
       rate(journalbeat_libbeat_output_events{type="acked"}[1m]) == 0
+  - alert: HighIngressCPU
+    labels: *infraticket
+    annotations:
+      message: |
+        We're seeing elevated CPU utilisation on the ingress nodes.
+    expr: |
+      aws_ecs_cpuutilization_average{cluster_name="prod-ingress", service_name="prod-frontend-v2"} > 70
```
CharlesIC: Isn't 70 too low as we discussed? Also this average in `aws_ecs_cpuutilization_average ` is over what period of time?

Krenair: Oh, right, did we say 90% in the end?
Isn't that just an average of the different tasks CPU utilisation at a point in time, not averaging together something over a period of time?

PR #390, opened 2020-03-31
Alert about problems connecting from ALB to frontend

Krenair:


terraform/modules/hub/files/prometheus/alerts.yml
```diff
@@ -174,5 +174,13 @@ groups:
     expr: |
       aws_ecs_cpuutilization_average{cluster_name="prod-ingress", service_name="prod-frontend-v2"} > 80
     for: 5m
+  - alert: FrontendConnectionErrors
+    labels: *infraticket
+    annotations:
+      message: |
+        The frontend ALB is having difficulty connecting to the service in ECS.
+    expr: |
+      aws_applicationelb_targetconnectionerrorcount_sum{load_balancer =~ "app/prod-ingress/.*"} > 0
```
Krenair, line 183: hm we probably shouldn't hardcode prod here, should probably fix it above too. also we probably need to ensure we get the -ingress-frontend target group?

blairboy362:
```suggestion
      aws_applicationelb_targetconnectionerrorcount_sum{load_balancer =~ "app/.*-ingress/.*", target_group =~ "targetgroup/.*-ingress-frontend/.*"} > 0
```

blairboy362: As a wild guess...

PR #393, opened 2020-04-02
Detect device identifier of prometheus EBS volume for mounting.

blairboy362:
EBS volumes are mapped to device identifiers in the kernel in a way that
is unpredictable (it gives numbers and letters based on the order in which
it becomes aware of them and so it cannot be relied upon). We are seeing
an issue whereby the ordering of the devices sometimes reverses and so an
attempt is made to mount the wrong device.

This change attempts to discover the identifier by looking it up by disk
size. It's a hack, but, for the moment, prometheus is only using a single
disk of that size.


blairboy362:
Closing this as it's all gone a bit weird following Github's issues this evening.

PR #397, opened 2020-04-14
Try running apt forever in ECS cloud-init

blairboy362:
We keep seeing issues during EC2 instance cloud-init where APT fails to
run successfully because it can't access a lock it needs. Instead of
checking for the locks ourselves and still hitting a potential race
condition, swtich the logic so we just keep retrying on failure. The end
result may be the same (an EC2 instance that fails to join the ECS
cluster) but this way there's more chance of it eventually succeeding.

### DIFF
--- a/terraform/modules/hub/modules/ecs_asg/files/cloud-init.sh
+++ b/terraform/modules/hub/modules/ecs_asg/files/cloud-init.sh
@@ -3,34 +3,12 @@ set -ueo pipefail
 
 export DEBIAN_FRONTEND=noninteractive
 
-function wait-for-lock() {
-	lock_file=$1
-	while fuser "$lock_file" >/dev/null 2>&1 ; do
-		echo "Waiting for $lock_file to be released..."
-		sleep 3
-	done
-}
-
-function check-dpkg-lock-before-use() {
-  status=1
-  for (( retry=0; retry<6 && status!=0; retry++ ))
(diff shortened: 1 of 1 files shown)


blairboy362:
I should also add that I did some snowflake -style testing with Staging to verify the behaviour of this change:

```
E: Could not get lock /var/lib/dpkg/lock-frontend - open (11: Resource temporarily unavailable)
E: Unable to acquire the dpkg frontend lock (/var/lib/dpkg/lock-frontend), is another process using it?
Executing apt-get dist-upgrade --yes failed. Sleeping...
E: Could not get lock /var/lib/dpkg/lock-frontend - open (11: Resource temporarily unavailable)
E: Unable to acquire the dpkg frontend lock (/var/lib/dpkg/lock-frontend), is another process using it?
Executing apt-get dist-upgrade --yes failed. Sleeping...
Reading package lists... Done
Building dependency tree
Reading state information... Done
Calculating upgrade... Done
0 upgraded, 0 newly installed, 0 to remove and 0 not upgraded.
```

blairboy362:
@galund There are many possible reasons APT will fail. But the end result, and reporting mechanism back to us, is no worse than it was before. Cloud-init is "fire-and-forget" script execution, so in both cases (pre- and post- this change) we just end up with an EC2 instance that can't be used by ECS.

PR #399, opened 2020-04-15
BAU: Remove now redundant provisioner from cognito in self-service

jakubmiarka:
The provisioner was needed to set the TOTP MFA since terraform did no support
setting it. The latest version of the AWS provider now supports it so it
can be set and doesn't need the provisioner.


j-maynard:
🦁

PR #400, opened 2020-04-16
Improve stability of docker state during ECS EC2 initialisation.

blairboy362:
We occasionally see issues where tasks can't start on an instance. The port
mapping required by the task can't be satisfied as the port is already
bound on the host. The working hypothesis is that this is due to docker
getting into a bad state as a result of the reboot at the end of the
cloud-init script leaving port mappings behind. These port mappings get
restored on reboot by docker before any tasks are (re-)registered by ECS.

By moving control of the ECS daemon to systemd, and not running ecs during
cloud-init we should see a reduction in the recurrence of the docker
state problem.

### DIFF
--- a/terraform/modules/hub/modules/ecs_asg/files/cloud-init.sh
+++ b/terraform/modules/hub/modules/ecs_asg/files/cloud-init.sh
@@ -153,22 +153,48 @@ iptables -t nat -A PREROUTING -p tcp -d 169.254.170.2 --dport 80 -j DNAT --to-de
 iptables -t nat -A OUTPUT -d 169.254.170.2 -p tcp -m tcp --dport 80 -j REDIRECT --to-ports 51679
 iptables-save > /etc/iptables/rules.v4
 
+# Have systemd manage the ecs agent
+# We have occasionally seen issues where docker gets into a bad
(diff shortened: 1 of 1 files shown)


blairboy362:
I've tested the systemd stuff on an instance in Staging to make sure it comes up correctly following a reboot etc. etc.

terraform/modules/hub/modules/ecs_asg/files/cloud-init.sh
```diff
@@ -190,6 +216,15 @@ docker run \
...
   ${ecs_agent_image_identifier}
 
+ExecStop=/usr/bin/docker stop -t 120 ecs-agent
+
+[Install]
+WantedBy=multi-user.target
+EOF
+
+systemctl daemon-reload
+systemctl enable ecs
```
adityapahuja, line 226:
```suggestion
systemctl enable ecs
systemctl start ecs
```

Can we start ecs service immediately?

blairboy362: See the commit message and the PR description. The whole point of this change is to not start the ECS container during clout-init (i.e. until after the reboot that happens at the bottom of the script).

PR #401, opened 2020-04-16
Independently scale-out frontend, analytics, and metadata 

chrisfarms:
## What

Allow us to set the scale of the `number_of_frontend_apps`, `number_of_analytics_apps` and `number_of_metadata_apps` independently from the global `number_of_apps`

## Why

So that we can scale out only the application that requires scaling out and avoid needing to overprovision the ECS cluster size.

## Context

frontend, analytics and metadata apps all live on the same "ingress" ECS cluster with a fixed cluster size, making it tricky to scale out the apps while keeping track of how big the cluster should be.

This changes uses [asg capacity provider](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/cluster-auto-scaling.html#asg-capacity-providers) to automatically scale the "ingress" cluster size based on the desired number of frontend, analytics and metadata app instances configured.

to ensure that two of the same task don't run on the same instance, we use the "spread" placement strategy.

## :warning: Before Merging

* [x] Ensure https://github.com/alphagov/verify-infrastructure-config/pull/285 has been merged FIRST to avoid accidentally scaling down frontend instances in prod!

### DIFF
--- a/terraform/modules/hub/ingress.tf
+++ b/terraform/modules/hub/ingress.tf
@@ -290,14 +290,15 @@ resource "aws_route53_record" "ingress_www" {
 module "ingress_ecs_asg" {
   source = "./modules/ecs_asg"
 
-  ami_id              = data.aws_ami.ubuntu_bionic.id
-  deployment          = var.deployment
-  cluster             = "ingress"
-  vpc_id              = aws_vpc.hub.id
-  instance_subnets    = aws_subnet.internal.*.id
-  number_of_instances = var.number_of_apps * 2
-  domain              = local.root_domain
-  instance_type       = var.ingress_instance_type
+  ami_id           = data.aws_ami.ubuntu_bionic.id
+  deployment       = var.deployment
+  cluster          = "ingress"
+  vpc_id           = aws_vpc.hub.id
+  instance_subnets = aws_subnet.internal.*.id
+  min_size         = var.number_of_apps * 2
+  max_size         = var.number_of_apps * 2
+  domain           = local.root_domain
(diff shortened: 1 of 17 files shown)


terraform/modules/hub/ingress.tf
```diff
@@ -290,14 +290,15 @@ resource "aws_route53_record" "ingress_www" {
...
-  number_of_instances = var.number_of_apps * 2
-  domain              = local.root_domain
-  instance_type       = var.ingress_instance_type
+  ami_id           = data.aws_ami.ubuntu_bionic.id
+  deployment       = var.deployment
+  cluster          = "ingress"
+  vpc_id           = aws_vpc.hub.id
+  instance_subnets = aws_subnet.internal.*.id
+  min_size         = 2
+  max_size         = 20
```
blairboy362, line 299: Not sure I understand why this is `20` and not `var.number_of_apps * 2`

chrisfarms: 1. number_of_apps is the number of tasks for each of the non-{frontend,metadata,analytics} apps ... so number_of_apps is not related to anything in the ingress cluster
2. 20 is a somewhat arbitrary upper limit on the number of nodes in the cluster... we currently run about 5 nodes in that cluster in prod, so I just picked an upper limit 4x that ... the actual number of nodes will be determined by the resources required to hold number_of_frontend_app + number_of_metadata_apps + number_of_analytics_apps all with placement "spread"

blairboy362: That makes sense. Thanks.

PR #404, opened 2020-04-20
Revert auto ingress scaling

chrisfarms:
reverts...

* https://github.com/alphagov/verify-infrastructure/pull/401
* https://github.com/alphagov/verify-infrastructure/pull/402
* https://github.com/alphagov/verify-infrastructure/pull/403

as it's disruptive (causes some downtime during deploy) and we're not in the mood for disruptive right now


chrisfarms:
hmm missing some... one sec

PR #406, opened 2020-05-04
add aws cli and some friends to the awsc image

chrisfarms:
we need the aws cli in the awsc concourse-task image so that we can implement some
logic waiting for ECS-tasks to become ready, jq is included to make
scripting of aws cli calls easier


Krenair:
For use in https://github.com/alphagov/verify-infrastructure-config/pull/302

PR #407, opened 2020-05-07
Upgrade squid to 4.11-r0

blairboy362:
I reviewed the changelog and nothing jumps out as affecting us so we should be safe. It'll be obvious if something does catastrophically break and is easily reverted.

### DIFF
--- a/dockerfiles/squid/Dockerfile
+++ b/dockerfiles/squid/Dockerfile
@@ -1,6 +1,6 @@
-FROM alpine:3.8
+FROM alpine:3.11
 
-RUN apk add --no-cache squid=3.5.27-r4
+RUN apk add --no-cache squid=4.10-r0
 
 RUN    touch /etc/squid/squid.conf \
     && chown squid /etc/squid/squid.conf


dockerfiles/squid/Dockerfile
```diff
@@ -1,6 +1,6 @@
 FROM alpine:3.8
```
adityapahuja, line 1:
```suggestion
FROM alpine:3.11
```

Alpine 3.11 has Squid v4.11-r0 package.

adityapahuja: I mean Squid v4.10-r0.

adityapahuja: When I run docker build on this file, I get the following error message.

```
fetch http://dl-cdn.alpinelinux.org/alpine/v3.8/community/x86_64/APKINDEX.tar.gz
ERROR: unsatisfiable constraints:
  squid-3.5.27-r4:
    breaks: world[squid=4.11-r0]
The command '/bin/sh -c apk add --no-cache squid=4.11-r0' returned a non-zero code: 1
```

dockerfiles/squid/Dockerfile
```diff
@@ -1,6 +1,6 @@
...
 
-RUN apk add --no-cache squid=3.5.27-r4
+RUN apk add --no-cache squid=4.11-r0
```
adityapahuja, line 3:
```suggestion
RUN apk add --no-cache squid=4.10-r0
```
It works if we use alpine:3.11.

dockerfiles/squid/Dockerfile
blairboy362: Thanks @adityapahuja. I clearly ran my test before saving the changes...way too tired.

dockerfiles/squid/Dockerfile
blairboy362: Fixed.

dockerfiles/squid/Dockerfile
adityapahuja: 👍

PR #411, opened 2020-06-04
Configure a copy of the config service to run on Fargate

Krenair:
You may want to use `diff -rb terraform/modules/hub/modules/ecs_app terraform/modules/hub/modules/ecs_fargate_app` to review part of this PR. By and large it's the existing thing with some tweaks to send it to Fargate instead of EC2.

### DIFF
--- a/terraform/modules/hub/modules/ecs_fargate_app/ecs.tf
+++ b/terraform/modules/hub/modules/ecs_fargate_app/ecs.tf
@@ -1,0 +1,41 @@
+module "cluster_ecs_roles" {
+  source = "../ecs_iam_role_pair"
+
+  deployment       = var.deployment
+  service_name     = var.cluster
+  tools_account_id = var.tools_account_id
+  image_name       = var.image_name
+}
+
+resource "aws_ecs_task_definition" "cluster" {
+  family                   = local.identifier
+  container_definitions    = var.task_definition
+  execution_role_arn       = module.cluster_ecs_roles.execution_role_arn
+  task_role_arn            = module.cluster_ecs_roles.task_role_arn
+  network_mode             = "awsvpc"
+  requires_compatibilities = ["FARGATE"]
+  cpu                      = var.cpu
+  memory                   = var.memory
+}
+
+output "task_role_name" {
+  value = module.cluster_ecs_roles.task_role_name
+}
+
+resource "aws_ecs_service" "cluster" {
+  name            = local.identifier
+  cluster         = var.ecs_cluster_id
+  task_definition = aws_ecs_task_definition.cluster.arn
+
+  desired_count                      = var.number_of_tasks
+  deployment_minimum_healthy_percent = var.deployment_min_healthy_percent
+  deployment_maximum_percent         = var.deployment_max_percent
+
+  launch_type = "FARGATE"
+
+  load_balancer {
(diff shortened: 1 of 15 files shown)


terraform/modules/hub/modules/ecs_fargate_app/variables.tf
```diff
@@ -0,0 +1,66 @@
+variable "cluster" {}
```
blairboy362, line 1: Maybe rename this to `service`

terraform/modules/hub/modules/ecs_fargate_app/ecs.tf
```diff
@@ -0,0 +1,41 @@
+module "cluster_ecs_roles" {
+  source = "../ecs_iam_role_pair"
+
+  deployment       = var.deployment
+  service_name     = var.cluster
+  tools_account_id = var.tools_account_id
+  image_name       = var.image_name
+}
+
+resource "aws_ecs_task_definition" "cluster" {
```
blairboy362, line 10: `cluster` feels wrong for the name here. Maybe 'service' or `app`? Not sure...

terraform/modules/hub/modules/ecs_fargate_app/ecs.tf
```diff
@@ -0,0 +1,41 @@
...
+  requires_compatibilities = ["FARGATE"]
+  cpu                      = var.cpu
+  memory                   = var.memory
+}
+
+output "task_role_name" {
+  value = module.cluster_ecs_roles.task_role_name
+}
+
+resource "aws_ecs_service" "cluster" {
```
blairboy362, line 25: `service` or `app` or something?

terraform/modules/hub/modules/ecs_fargate_app/lb.tf
```diff
@@ -0,0 +1,63 @@
+resource "aws_lb" "cluster" {
```
blairboy362, line 1: `service` or `app`?

terraform/modules/hub/modules/ecs_fargate_app/route53.tf
```diff
@@ -0,0 +1,23 @@
+resource "aws_route53_zone" "cluster" {
```
blairboy362, line 1: Is this needed?

terraform/modules/hub/modules/ecs_fargate_app/security_groups.tf
```diff
@@ -0,0 +1,33 @@
...
+  description = "${local.identifier}-lb"
+
+  vpc_id = var.vpc_id
+}
+
+output "lb_sg_id" {
+  value = aws_security_group.lb.id
+}
+
+resource "aws_security_group_rule" "task_ingress_from_lb" {
```
blairboy362, line 12: can `microservice_connection` be used in place of these two security group rules?

terraform/modules/hub/modules/ecs_fargate_app/route53.tf
Krenair: Maybe. Everything in this module is a copy of the equivalent from ecs_app, and unless there's a Fargate-specific reason to change it I'd prefer to keep it the same for now.

PR #424, opened 2020-06-15
Config on Fargate: Try sending logs to Logit

Krenair:
This is experimental and may not work.

### DIFF
--- a/terraform/modules/hub/files/tasks/hub-config-fargate.json
+++ b/terraform/modules/hub/files/tasks/hub-config-fargate.json
@@ -1,5 +1,23 @@
 [
   {
+    "essential": true,
+    "image": "906394416424.dkr.ecr.eu-west-2.amazonaws.com/aws-for-fluent-bit:latest",
+    "name": "log_router",
+    "firelensConfiguration": {
+      "type": "fluentbit"
+    },
+    "logConfiguration": {
+      "logDriver": "awslogs",
+      "options": {
+        "awslogs-group": "${deployment}-hub",
+        "awslogs-region": "${region}",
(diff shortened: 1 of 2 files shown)


terraform/modules/hub/files/tasks/hub-config-fargate.json
```diff
@@ -24,12 +42,12 @@
...
     "logConfiguration": {
-      "logDriver": "awslogs",
+      "logDriver": "awsfirelens",
       "options": {
-        "awslogs-group": "${deployment}-hub",
-        "awslogs-region": "${region}",
-        "awslogs-stream-prefix": "config-nginx",
-        "awslogs-create-group": "true"
+        "Name": "es",
+        "Host": "https://${logit_elasticsearch_url}:443",
```
adityapahuja, line 48: Do we need `:443` if we use `https://`?

adityapahuja: TLS support is currently disabled by default. Can we enable TLS support?

Krenair: This is just the string copied directly from the journalbeat config. We probably can get rid of the port, but I think TLS support is already enabled?

adityapahuja: By default, tls is set to off. Source: https://docs.fluentbit.io/manual/pipeline/outputs/tcp-and-tls

Krenair: Hm okay yes, we need to get rid of https:// and :443 here and set that.

PR #433, opened 2020-06-18
Ship logs from (almost) all remaining Hub components to Splunk

Krenair:
Except beat-exporter, which this work will presumably end in the termination of.


philandstuff:
Presumably this change will be noticable for verify 2nd line?  Have we set their expectations about this?

Krenair:
Probably and yes (well, technically hub-and-eidas). We're going to be pausing the prod pipeline and looking at this in staging.

PR #445, opened 2020-06-24
Revert egress proxy changes around the saml-engine ECS agent, replace with VPC Endpoint

Krenair:
This reverts commit 4095faa79c1ed2e6de583c435975ba00eff48b5b.
This reverts commit 2311f8f66b68791900bbf1177edba6d728572519.

It turns out to do that we would need to safelist a bunch of things in the egress
proxy such as ECS, ECS-A (Agent?), ECS-T (Telemetry?), SSM, and probably more
like ECR.


Krenair:
In retrospect this is what we probably should've done in the first place, had I known better what ECS agent would expect to talk to and what the existing restrictions around communications from saml-engine were.

PR #447, opened 2020-07-01
Upgrade prometheus AMI to Focal

philandstuff:
This upgrades the Prometheus AMI to focal fossa, the latest Ubuntu LTS
release.

This only upgrades Prometheus; we might want to upgrade the rest of
the ECS nodes, or they might become obsoleted by fargate.

I suspect there isn't much to do but it will probably be worth pausing
production to make sure there aren't any issues in staging when it
gets deployed.


philandstuff:
opened as draft because I'm not around to guide this through the pipeline.

PR #449, opened 2020-07-14
matomo: match fargate task size with task def

chrisfarms:
requested fargate task limits must match container requests for
cpu/memory.

### DIFF
--- a/terraform/modules/hub/files/matomo/matomo-task-def.json
+++ b/terraform/modules/hub/files/matomo/matomo-task-def.json
@@ -1,7 +1,7 @@
 [
   {
     "cpu": 3072,
-    "memory": 12288,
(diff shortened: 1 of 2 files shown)


philandstuff:
should this be configurable so that we can have a bigger one in prod?

chrisfarms:
@philandstuff Yes ... just trying to clear the broken deploy pipeline for now tho

adityapahuja:
> should this be configurable so that we can have a bigger one in prod?

I agree. We need to be able to scale up Matomo to avoid 5xx errors and also scale down when Matomo is not under pressure.

chrisfarms:
I do agree .. but we can address that later ... right now this is just repairing an issue with the terraform that is blocking the deploy pipeline.

PR #450, opened 2020-07-14
matomo: deploy as fargate service and allow configuring scale-out

chrisfarms:
* Adds var to set number of instances (default to 1 for now)
* Adds correct launch type


adityapahuja:
Can we have at least 2 instances so that we don't see Matomo downtime (ZDD)?

PR #463, opened 2020-07-28
Remove various security rules pointing at the old removed config EC2 service

Krenair:


Krenair:
it's not gonna be enough... just spotted prometheus_can_talk_to_config in terraform/modules/hub/prometheus.tf

PR #466, opened 2020-07-29
BAU allow log level to be set in config fargate task

idrop:
Allow value in terraform `var.hub_config_log_level` to be set as log level in the new fargate task.
This was set in the 'old' config task, which has been replaced by a fargated version.


Krenair:
you mean config?

PR #481, opened 2020-08-03
AUT-854: Set up SAML Proxy in Fargate

adityapahuja:
This change sets up a new service called SAML Proxy Fargate. This service will run in Fargate alongside with SAML Proxy running on EC2 instances.

Author: @adityapahuja

PR #490, opened 2020-08-04
Revert "AUT-854: Set up SAML Proxy in Fargate"

philandstuff:
Reverts alphagov/verify-infrastructure#481


philandstuff:
The new saml-proxy is having trouble sending to the event system; reverting while we debug.

philandstuff:
this was reverting the wrong PR, see #491 instead

PR #498, opened 2020-08-06
AUT-817: Set up SAML SOAP Proxy on Fargate

adityapahuja:
This change sets up a new service called SAML SOAP Proxy Fargate. It will run on Fargate alongside with SAML SOAP Proxy running on EC2 instances.

Author: @adityapahuja

PR #499, opened 2020-08-07
Revert "AUT-817: Set up SAML SOAP Proxy on Fargate"

adityapahuja:
Reverts alphagov/verify-infrastructure#498 until we sort name's 32 character limit issue.


adityapahuja:
This pull request is not needed. Found a solution to this problem (https://github.com/alphagov/verify-infrastructure/pull/500).

PR #500, opened 2020-08-07
AUT-817: Shorten the name of lb in integration environment

adityapahuja:
This change adds a special case to rename from "integration-saml-soap-proxy-fargate" to "integration-saml-soap-proxy-fg" to avoid the following error message.

`Error: "name" cannot be longer than 32 characters: "integration-saml-soap-proxy-fargate"`

Author: @adityapahuja


adityapahuja:
I have made one more change and ran terraform plan. It looks good. No more "name" cannot be longer than 32 characters" errors.

PR #503, opened 2020-08-11
Create ECS Fargate cluster for static ingress

blairboy362:
Includes a duplication of the whole stack due to client-ip requirements. A
future commit will remap the EIPs to the new LB (likely triggering some
downtime).

### DIFF
--- a/terraform/modules/hub/static_ingress.tf
+++ b/terraform/modules/hub/static_ingress.tf
@@ -251,3 +251,280 @@ resource "aws_lb_listener" "static_ingress_https" {
     target_group_arn = aws_lb_target_group.static_ingress_https.arn
   }
 }
+
+resource "aws_lb" "static_ingress_fargate" {
+  name                             = "${var.deployment}-static-ingress-fargate"
+  load_balancer_type               = "network"
+  internal                         = false
+  enable_cross_zone_load_balancing = true
+  subnets                          = aws_subnet.ingress.*.id
+}
+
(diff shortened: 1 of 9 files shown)


blairboy362:
I've updated the fargate task definitions to bring them inline with what is [allowed in fargate](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/AWS_Fargate.html). When the EC2 instances are removed I'll update the task definition accordingly.

chrisfarms (approved):

terraform/modules/hub/static_ingress.tf
```diff
@@ -251,3 +251,280 @@ resource "aws_lb_listener" "static_ingress_https" {
     target_group_arn = aws_lb_target_group.static_ingress_https.arn
   }
 }
+
+resource "aws_lb" "static_ingress_fargate" {
+  name                             = "${var.deployment}-static-ingress-fargate"
+  load_balancer_type               = "network"
+  internal                         = false
+  enable_cross_zone_load_balancing = true
+  subnets                          = [aws_subnet.public.*.id]
```
chrisfarms, line 260: `aws_subnet.public` resource doesn't seem to exist

blairboy362: Good spot! Should be `aws_subnet.ingress.*.id`. Updated and squashed.

chrisfarms (approved):

Krenair (approved):

PR #504, opened 2020-08-12
analytics: add analytics task to fargate cluster

chrisfarms:
deploy second set of analytics tasks into the shared fargate ecs
cluster.

the tasks are connected to the existing target group so will serve
traffic in parallel with the existing EC2 tasks.

Due to fargate cpu/memory constraints the task definition cpu requests
are lowered a little to fit.

The memory requests are left as is for now as they are tuned to fit a
certain number of tasks onto the EC2 instances. A follow up PR will
alter these values to use the maximum memory for the task once
EC2-service is removed.

### DIFF
--- a/terraform/modules/hub/analytics.tf
+++ b/terraform/modules/hub/analytics.tf
@@ -90,3 +90,65 @@ resource "aws_ecs_service" "analytics" {
     ]
   }
 }
+
+resource "aws_ecs_task_definition" "analytics_fargate" {
+  family                   = "${var.deployment}-analytics-fargate"
+  container_definitions    = data.template_file.analytics_task_def.rendered
+  execution_role_arn       = module.analytics_ecs_roles.execution_role_arn
+  network_mode             = "awsvpc"
+  requires_compatibilities = ["FARGATE"]
+  cpu                      = 256
+  memory                   = 512
+}
+
+resource "aws_ecs_service" "analytics_fargate" {
+  name            = "${var.deployment}-analytics"
+  cluster         = aws_ecs_cluster.fargate-ecs-cluster.id
+  task_definition = aws_ecs_task_definition.analytics_fargate.arn
+
+  desired_count                      = var.number_of_apps
+  deployment_minimum_healthy_percent = 50
+  deployment_maximum_percent         = 100
+
+  load_balancer {
+    target_group_arn = aws_lb_target_group.ingress_analytics.arn
+    container_name   = "nginx"
+    container_port   = "8443"
+  }
+
+  network_configuration {
+    subnets = aws_subnet.internal.*.id
+
+    security_groups = [
+      aws_security_group.analytics_task.id,
+      aws_security_group.hub_fargate_microservice.id,
+      aws_security_group.can_connect_to_container_vpc_endpoint.id,
+    ]
+  }
+
+  service_registries {
+    registry_arn = aws_service_discovery_service.analytics_fargate.arn
+    port         = 8443
+  }
+}
+
+resource "aws_service_discovery_service" "analytics_fargate" {
+  name = "${var.deployment}-analytics"
+
+  description = "service discovery for ${var.deployment}-analytics-fargate instances"
+
+  dns_config {
+    namespace_id = aws_service_discovery_private_dns_namespace.hub_apps.id
+
+    dns_records {
+      ttl  = 60
+      type = "SRV"
+    }
+
+    routing_policy = "MULTIVALUE"
+  }
+
+  health_check_custom_config {
+    failure_threshold = 2
+  }
+}
--- a/terraform/modules/hub/files/tasks/analytics.json
+++ b/terraform/modules/hub/files/tasks/analytics.json
@@ -2,7 +2,7 @@
   {
     "name": "nginx",
     "image": "${nginx_image_identifier}",
-    "cpu": 298,
+    "cpu": 256,
     "memory": 250,
     "essential": true,
     "portMappings": [


terraform/modules/hub/analytics.tf
```diff
@@ -90,3 +90,64 @@ resource "aws_ecs_service" "analytics" {
     ]
   }
 }
+
+resource "aws_ecs_task_definition" "analytics_fargate" {
+  family                   = "${var.deployment}-analytics"
```
blairboy362, line 95:
```suggestion
  family                   = "${var.deployment}-analytics-fargate"
```

terraform/modules/hub/analytics.tf
```diff
@@ -90,3 +90,64 @@ resource "aws_ecs_service" "analytics" {
...
+    target_group_arn = aws_lb_target_group.ingress_analytics.arn
+    container_name   = "nginx"
+    container_port   = "8443"
+  }
+
+  network_configuration {
+    subnets = aws_subnet.internal.*.id
+
+    security_groups = [
+      aws_security_group.analytics_task.id,
```
blairboy362, line 123: Not essential, but might be one of those things that gets forgotten about later.
```suggestion
      aws_security_group.analytics_task.id,
      aws_security_group.hub_fargate_microservice.id,
```

blairboy362: Oh, my understanding was `hub_fargate_microservice` basically meant that prometheus could connect to it

chrisfarms: ah ok, yeah I guess it makes sense to add it then, just in case it gets instrumented with a metrics endpoint in future :+1:

blairboy362:
@chrisfarms Do we need to add security group rules to `static_ingress.tf` to allow connection to the new analytics fargate task?

chrisfarms:
> @chrisfarms Do we need to add security group rules to `static_ingress.tf` to allow connection to the new analytics fargate task?

IIUC the routing is:

```
NLB(static_ingress) → HAProxy(static_ingress) → ALB(ingress) → microservice(analytics)
```

So I don't expect to have to change anything todo with static_ingress here since I haven't touched the `ALB(ingress)` (other than adding another task to target group)

If I'm wrong about that setup then you could be right.

blairboy362:
@chrisfarms You're right, I think. I was tying myself in knots over this stuff yesterday. I blame the heat frying my brain.

terraform/modules/hub/analytics.tf
```diff
@@ -90,3 +90,65 @@ resource "aws_ecs_service" "analytics" {
...
+      aws_security_group.hub_fargate_microservice.id,
+      aws_security_group.can_connect_to_container_vpc_endpoint.id,
+    ]
+  }
+
+  service_registries {
+    registry_arn = aws_service_discovery_service.analytics_fargate.arn
+    port         = 8443
+  }
```
philandstuff, line 132:
```suggestion
```

philandstuff: prometheus doesn't scrape analytics, there's no metrics endpoint for you here, move along

terraform/modules/hub/analytics.tf
```diff
@@ -90,3 +90,65 @@ resource "aws_ecs_service" "analytics" {
...
+      type = "SRV"
+    }
+
+    routing_policy = "MULTIVALUE"
+  }
+
+  health_check_custom_config {
+    failure_threshold = 2
+  }
+}
```
philandstuff, line 154:
```suggestion

```

chrisfarms:
@philandstuff yes @blairboy362 and I discussed this ... the nginx isn't scrapped, but adding service discovery later requires recreating the "service" so we thought better to put it in for consistency and not need it, rather than want it later and be in an annoying situation where we need to workaround downtime.

philandstuff:
Well can we at least not have the description

```
  description = "A service to allow Prometheus to discover ${var.deployment}-analytics instances"
```

as that is *bound* to cause confusion

terraform/modules/hub/analytics.tf
```diff
@@ -90,3 +90,65 @@ resource "aws_ecs_service" "analytics" {
...
+  name            = "${var.deployment}-analytics"
+  cluster         = aws_ecs_cluster.fargate-ecs-cluster.id
+  task_definition = aws_ecs_task_definition.analytics_fargate.arn
+
+  desired_count                      = var.number_of_apps
+  deployment_minimum_healthy_percent = 50
+  deployment_maximum_percent         = 100
+
+  load_balancer {
+    target_group_arn = aws_lb_target_group.ingress_analytics.arn
```
philandstuff, line 114: I don't remember if this is allowed... ISTR some horrible restriction that only one service could "own" a target group

philandstuff: after reading the docs, it looks like stuff has changed since I last looked at this. It's now possible for a single service to register with multiple target groups, so who knows it might also be possible for a single target group to host multiple services.

chrisfarms: I had this fear too... but I couldn't spot anything in docs, so was gonna try and see

chrisfarms:
removed the mention of "Prometheus" in description and amended the commit to explain that service disco is only there for consistency/future so that future travellers can feel safer removing if need be

terraform/modules/hub/analytics.tf
chrisfarms: replies at top-level PR comments ... gonna keep discovery enabled in case of future needs.

PR #509, opened 2020-08-14
frontend: run fargate task in parallel

chrisfarms:
create a new fargate deployment of frontend in parallel with the ec2
deployment and let it share the traffic load.

the split between two near identical "frontend.json" and
"frontend_xlarge.json" task definitions has been consolidated into a
single file with the requisite values tempalted out to facilitate
configuring the cpu/memory values to fit fargate constraints.

as we move away from the ec2 instances the "frontend_instance_type"
variable will become meaningless, so a future PR will tidy up all the
"if this instance type do that" stuff by making the cpu/memory of the
task configurable per environment.

This change gives slightly less cpu to prod frontend, however this should not be an issue since:

1. there will be twice as many instances running until the ec2 versions are disabled
1. frontend is pretty overprovisioned right now


terraform/modules/hub/hub_frontend.tf
```diff
@@ -128,6 +132,49 @@ resource "aws_ecs_service" "frontend_v2" {
   }
 }
 
+resource "aws_ecs_task_definition" "frontend_fargate" {
+  family                   = "${var.deployment}-frontend-fargate"
+  container_definitions    = data.template_file.frontend_task_def.rendered
+  network_mode             = "awsvpc"
+  execution_role_arn       = module.frontend_ecs_roles.execution_role_arn
+  requires_compatibilities = ["FARGATE"]
+  cpu                      = var.ingress_instance_type == "t3.xlarge" ? 2048 : 512
```
adityapahuja, line 141:
```suggestion
  cpu                      = var.ingress_instance_type == "t3.xlarge" ? 2048 : 1024
```

nginx_cpu = 256 + frontend_cpu = 768 = 1024

chrisfarms: oops :+1:

adityapahuja:
We will need to keep an eye on Frontend CPU usage in production.

PR #510, opened 2020-08-14
frontend: removing ec2-based frontend tasks

chrisfarms:
## What

* Fixes cloudwatch metric collection for analytics and frontend tasks
* Removes the container-level cpu/mem constaints to simplify configuration (these are only required for ec2-tasks)
* Removes the old ec2-tasks
* Changes scaling to be based on the number of frontend instances, instead of basing scaling on combination of the number_of_apps and the now redundent ingress_instance_type

## :warning: Before merging

Ensure https://github.com/alphagov/verify-infrastructure/pull/509 is merged and deployed to production first!!!: 

Deploy at same time as https://github.com/alphagov/verify-infrastructure-config/pull/354

### DIFF
--- a/terraform/modules/hub/hub_frontend.tf
+++ b/terraform/modules/hub/hub_frontend.tf
@@ -88,46 +88,6 @@ module "frontend_ecs_roles" {
   image_name       = "verify-frontend"
 }
 
-resource "aws_ecs_task_definition" "frontend" {
-  family                = "${var.deployment}-frontend"
-  container_definitions = data.template_file.frontend_task_def.rendered
-  network_mode          = "awsvpc"
-  execution_role_arn    = module.frontend_ecs_roles.execution_role_arn
-}
-
-# This is called frontend_v2 because there was an old frontend service
-# that coexisted at the same time for a while and life is sometimes
-# too short to `terraform state mv`
-resource "aws_ecs_service" "frontend_v2" {
-  name            = "${var.deployment}-frontend-v2"
-  cluster         = aws_ecs_cluster.ingress.id
-  task_definition = aws_ecs_task_definition.frontend.arn
-
-  desired_count                      = var.number_of_apps
-  deployment_minimum_healthy_percent = 50
-  deployment_maximum_percent         = 100
-
-  load_balancer {
-    target_group_arn = aws_lb_target_group.ingress_frontend.arn
-    container_name   = "nginx"
-    container_port   = "8443"
-  }
-
-  network_configuration {
-    subnets = aws_subnet.internal.*.id
-
-    security_groups = [
-      aws_security_group.frontend_task.id,
-      aws_security_group.can_connect_to_container_vpc_endpoint.id,
-    ]
-  }
-
-  service_registries {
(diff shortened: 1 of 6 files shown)


terraform/modules/hub/variables.tf
```diff
@@ -14,6 +14,11 @@ variable "number_of_apps" {
   default = 2
 }
 
+variable "number_of_frontend_apps" {
+  type    = number
+  default = 2
+}
```
adityapahuja, line 20: I wouldn't want to set number of frontend_apps to 2 in production by default.

philandstuff: https://github.com/alphagov/verify-infrastructure-config/pull/354 will set it to 16. I'm happy for the default to be 2.

terraform/modules/hub/hub_frontend.tf
```diff
@@ -92,62 +88,22 @@ module "frontend_ecs_roles" {
...
+  memory                   = 2048
 }
 
 resource "aws_ecs_service" "frontend_fargate" {
   name            = "${var.deployment}-frontend"
   cluster         = aws_ecs_cluster.fargate-ecs-cluster.id
   task_definition = aws_ecs_task_definition.frontend_fargate.arn
 
-  desired_count                      = var.number_of_apps
+  desired_count                      = var.number_of_frontend_apps
```
adityapahuja, line 106:
```suggestion
  desired_count                      = var.deployment != "prod" ? var.number_of_frontend_apps : 16
```

adityapahuja: This change is temporary and can be removed once the pull request is https://github.com/alphagov/verify-infrastructure-config/pull/354 is merged.

philandstuff: I would have thought we could just pause the resources in concourse to ensure the changes go out together

adityapahuja: It is better than pausing the resources in concourse.

chrisfarms: Id rather just pause and merge both PRs... Rather than have to get another PR merged after

chrisfarms: Merge the other PR first to reduce the risk if concerned ... That way it'll just error and not deploy to prod

adityapahuja: If you unpause the jobs in Concourse, will Concourse push the change (changing the number of frontend apps to 2) to production first and then push the next change (increase the number of frontend apps to 16) to production?

chrisfarms: The process I usually do for this kind of thing is:

* Pause "deploy-hub-staging" job
* Merge both PRs
* Click "refresh" on each of the git resources in concourse (to ensure they have latest version)
* Wait a little bit
* Unpause the "deploy-hub-staging" job
* Look at the build log in concourse to double check both of the git resources are using the expected version
* If not... pause prod and work out why, else all good :+1:

adityapahuja: Thanks. It makes sense.

PR #517, opened 2020-08-17
egress-proxy: deploy task to fargate cluster

chrisfarms:
creates a fargate egress-proxy task.

lowers container memory limit a little to fit in the 3072 fargate task
size.

we will be moving away from a "classic" ELB, so there's some prep for
that:

creates/updates security groups ready for either connecting via NLB or
directly (both of which will require security groups on the task to
allow ingress rather than on the load balancer since NLBs don't accept
security groups).

Adds service discovery for consitancy with other tasks and so that we
can connect directly.

This does not put anything "live" yet, just gets the task setup. A
follow up PR will direct traffic to the task once it's working.

### DIFF
--- a/terraform/modules/hub/egress_proxy.tf
+++ b/terraform/modules/hub/egress_proxy.tf
@@ -28,6 +28,18 @@ resource "aws_security_group" "egress_via_proxy" {
   vpc_id = aws_vpc.hub.id
 }
 
+resource "aws_security_group_rule" "egress_to_proxy_task" {
+  type     = "egress"
+  protocol = "tcp"
+
+  from_port = 8080
+  to_port   = 8080
+
+  # source is destination for egress rules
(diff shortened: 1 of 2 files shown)


terraform/modules/hub/egress_proxy.tf
```diff
@@ -28,6 +28,18 @@ resource "aws_security_group" "egress_via_proxy" {
   vpc_id = aws_vpc.hub.id
 }
 
+resource "aws_security_group_rule" "egress_to_proxy_task" {
```
philandstuff, line 31: I'd find it less confusing if this rule was next to the equivalent `ingress` rule below

chrisfarms: Can I move them around once I've deleted all the old stuff?

philandstuff: sure

PR #519, opened 2020-08-17
prometheus: remove logit shipping

chrisfarms:
The prometheus ECS cluster is not being migrated to Fargate.

All ECS container logs are being shipped to cloudwatch/splunk (including prometheus). 

Journal-beat is currently still shipping non-ECS container logs from the ECS EC2 instances (systemd unit logs etc) to Logit. We do not want to keep shipping things to Logit just for the auxiliary processes on the prometheus nodes. In the rare cases we need to investigate these other logs we can still use SSM to connect to the nodes and view the logs with `journalctl`

This change:

* removes the beat-exporter containers from the Prometheus cluster
* uninstalls the journal-beat processes from the nodes
* defaults the docker daemon to use the `awslogs` driver so that logs from anything running in Docker get shipped to the same log group as all other hub components.

### DIFF
--- a/terraform/modules/hub/files/cloud-init/prometheus.sh
+++ b/terraform/modules/hub/files/cloud-init/prometheus.sh
@@ -68,7 +68,7 @@ run-until-success apt-get install --yes docker.io
 cat <<EOF > /etc/systemd/system/docker.service.d/override.conf
 [Service]
 ExecStart=
-ExecStart=/usr/bin/dockerd --log-driver journald --dns 10.0.0.2
+ExecStart=/usr/bin/dockerd --log-driver awslogs --log-opt awslogs-region=eu-west2 --log-opt awslogs-group ${cloudwatch_log_group} --dns 10.0.0.2
 EOF
 
 # Reload systemctl daemon to pick up new override files
--- a/terraform/modules/hub/files/cloud-init/prometheus.sh
+++ b/terraform/modules/hub/files/cloud-init/prometheus.sh
@@ -68,7 +68,7 @@ run-until-success apt-get install --yes docker.io
 cat <<EOF > /etc/systemd/system/docker.service.d/override.conf
 [Service]
 ExecStart=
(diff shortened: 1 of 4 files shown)


terraform/modules/hub/files/cloud-init/prometheus.sh
```diff
@@ -68,59 +68,14 @@ run-until-success apt-get install --yes docker.io
 cat <<EOF > /etc/systemd/system/docker.service.d/override.conf
 [Service]
 ExecStart=
-ExecStart=/usr/bin/dockerd --log-driver journald --dns 10.0.0.2
+ExecStart=/usr/bin/dockerd --log-driver awslogs --log-opt awslogs-region=eu-west2 --log-opt awslogs-group ${cloudwatch_log_group} --dns 10.0.0.2
```
philandstuff, line 71:
```suggestion
ExecStart=/usr/bin/dockerd --log-driver awslogs --log-opt awslogs-region=eu-west-2 --log-opt awslogs-group ${cloudwatch_log_group} --dns 10.0.0.2
```
I don't know that this is right, but `eu-west2` looks wrong

terraform/modules/hub/files/cloud-init/prometheus.sh
```diff
@@ -68,59 +68,14 @@ run-until-success apt-get install --yes docker.io
...
-  hosts: ["https://${logit_elasticsearch_url}:443"]
-  headers:
-    Apikey: ${logit_api_key}
-EOF
-# It seems that journalbeat is very unhappy if docker isn't running
-# when it starts, so let's update the systemd unit to reflect that
-sed -i -e 's/Wants=.*/& docker.service/' /lib/systemd/system/journalbeat.service
-sed -i -e 's/After=.*/& docker.service/' /lib/systemd/system/journalbeat.service
-systemctl enable --now journalbeat
-
```
philandstuff, line 123: now that we've got rid of journalbeat, we could run prometheus on `m6g` instances (slight troll, but I think the issues I found in https://github.com/alphagov/prometheus-aws-configuration-beta/pull/400 are not a concern in this PR, except the only-two-AZs one)

PR #521, opened 2020-08-18
egress-proxy: switch to using the fargate version

chrisfarms:
We currently have two egress-proxy services deployed (one on ECS/EC2 and
one on ECS/Fargate).

The EC2 version uses a Classic ELB to forward traffic to tasks.

The fargate version has DNS service discovery enabled so that traffic can be
sent directly to the running tasks.

This changes the saml-engine tasks to use the addresses of the
discovered fargate tasks and removes the use of the egress proxy from
the remaining ec2 clusters (which both have open egress to 80/443 anyway).

### DIFF
--- a/terraform/modules/hub/modules/ecs_asg/files/cloud-init.sh
+++ b/terraform/modules/hub/modules/ecs_asg/files/cloud-init.sh
@@ -11,20 +11,8 @@ function run-until-success() {
   done
 }
 
-CURL="curl"
-if [ -n "${egress_proxy_url_with_protocol}" ]; then
-  CURL="curl --proxy ${egress_proxy_url_with_protocol}"
-fi
-
 # Apt
 echo 'Configuring apt'
-mkdir -p /etc/apt/apt.conf.d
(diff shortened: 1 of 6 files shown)


chrisfarms:
> can we make a point of testing the DNS caching thing in staging? ie deploy, kill the egress proxy tasks, run a test journey and make sure saml-engine still works

yep fully intend on doing as you say ... will pause deploy-prod/int ... poke it a bit before continuing :+1:

terraform/modules/hub/modules/ecs_asg/files/cloud-init.sh
```diff
@@ -11,20 +11,8 @@ function run-until-success() {
   done
 }
 
-CURL="curl"
```
Krenair, line 14: this is still in use down on line 91 in the new version

chrisfarms: nice spot :+1: ... fixed

paroxp (approved):

PR #524, opened 2020-08-18
prometheus: fix log group iam

chrisfarms:
## What

Grant full cloudwatch logs access to the iam profile in the prometheus ec2 nodes


philandstuff:
note: all the apps currently have `"*"` permissions. we should review if this is right, but right now this is to get prometheus running again.

PR #527, opened 2020-08-19
Fix ECS task definition for static ingress on fargate

blairboy362:
Hopefully this duplication will be short lived.

### DIFF
--- a/terraform/modules/hub/files/tasks/static-ingress-fargate.json
+++ b/terraform/modules/hub/files/tasks/static-ingress-fargate.json
@@ -1,0 +1,43 @@
+[
+  {
+    "name": "static-ingress",
+    "image": "${image_identifier}",
+    "cpu": ${allocated_cpu},
+    "memory": ${allocated_memory},
+    "essential": true,
+    "portMappings": [
+      {
+        "containerPort": ${bind_port},
+        "hostPort": ${bind_port}
+      }
+    ],
+    "environment": [{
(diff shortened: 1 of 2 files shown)


blairboy362:
It's a duplicate of the non-fargate container definition but with the `systemControls` object removed.

Krenair:
looks like you actually removed `linuxParameters`?

blairboy362:
Oh dear. Trust the code, not the comment :) `linuxParameters` was what I intended to remove.

Krenair (approved):
makes sense from the errors, don't think we should need any special linux permissions for this?

terraform/modules/hub/files/tasks/static-ingress-fargate.json
```diff
@@ -0,0 +1,43 @@
+[
+  {
+    "name": "static-ingress",
+    "image": "${image_identifier}",
+    "cpu": ${allocated_cpu},
+    "memory": ${allocated_memory},
```
chrisfarms, line 6: non-blocking: these are optional in fargate as the memory is set at the task-level

chrisfarms (approved):

terraform/modules/hub/files/tasks/static-ingress-fargate.json
blairboy362: Yeah. It's on my list to tidy these up once the migration to fargate is done.

PR #529, opened 2020-08-19
Stop haproxy from trying to bind to 0.0.0.0

blairboy362:
In a fargate world binding to 0.0.0.0 is not allowed. Hopefully this is
the fix.


Krenair (approved):

chrisfarms (approved):
I thought that was the same as the deafult :man_shrugging: give it a go

blairboy362:
We don't think this is the problem/solution. Going to try something else.

PR #535, opened 2020-08-20
scaling: scale-in and scale-down prod apps

chrisfarms:
## What

* Make vertical scale of applications consistent accross environments. 
* Simplify configuration by removing unnecessary resource limits
* Allow scaling-out each micro-service independently 

## Why

Currently the "vertical" size of containers in prod differs from the size in staging/integration, for example the "frontend" tasks run in 2vcpu 8GB container in production, and run in 2vcpu 4GB containers in staging. This complicates the deployment configuration with conditional statements, and means that the applications may behave significantly differently between environments, which can lead to problems being missed in staging.

This disparity came about because there was a need to scale verify, but due to the way ECS-on-EC2 was setup, it was easier to scale-up (bigger instances) than it was to scale-out (more instances). However now we have moved the micro-services to ECS-on-Fargate, no such complication exists and we can now scale-out (more instances) each application individually as desired.

## :warning: Merge in sync with...

* https://github.com/alphagov/verify-infrastructure-config/pull/363

### DIFF
--- a/terraform/modules/hub/variables.tf
+++ b/terraform/modules/hub/variables.tf
@@ -73,11 +73,11 @@ variable "prometheus_volume_size" {
 }
 
 variable "publically_accessible_from_cidrs" {
-  type = "list"
+  type = list
 }
 
 variable "mgmt_accessible_from_cidrs" {
-  type = "list"
+  type = list
 }
 
 variable "redis_cache_size" {
--- a/terraform/modules/hub/variables.tf
+++ b/terraform/modules/hub/variables.tf
@@ -160,26 +160,6 @@ variable "jvm_options" {
   default = "-XX:InitialRAMPercentage=50 -XX:MaxRAMPercentage=80"
 }
 
-variable "config_memory_hard_limit" {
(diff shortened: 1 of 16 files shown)


terraform/modules/hub/variables.tf
```diff
@@ -19,6 +19,51 @@ variable "number_of_frontend_apps" {
...
+  type    = number
+  default = 2
+}
+
+variable "number_of_egress_proxy_apps" {
+  type    = number
+  default = 2
+}
+
+variable "number_of_static_ingress_apps" {
```
blairboy362, line 62: It doesn't look like you've used this anywhere. Is this a placeholder for when the static-ingress is migrated?

blairboy362 (approved):

terraform/modules/hub/variables.tf
chrisfarms: yeah exactly ... this means don't have to do a "sync up" deploy of both repos later

philandstuff (approved):

PR #543, opened 2020-08-24
[hub] restore container-level memory limits

philandstuff:
These were removed in ad2410f.  However we have a theory that this
removal is causing the JVM to be unable to regulate its memory usage.
We use the options `-XX:InitialRAMPercentage=50
-XX:MaxRAMPercentage=80` to specify JVM RAM usage as a percentage of
available RAM, but the removal of these limits seems to have confused
the JVM about how much RAM is really available.

There are also similar reports on the AWS forums:
https://forums.aws.amazon.com/thread.jspa?threadID=297114

This sets each of the app nginxes to have 250 MiB of memory, and each
of the apps to have the rest of the task memory.  Happily, the new
`templatefile` function lets us do maths in the template to make this
easier to understand.

### DIFF
--- a/terraform/modules/hub/files/tasks/hub-config.json
+++ b/terraform/modules/hub/files/tasks/hub-config.json
@@ -2,6 +2,7 @@
   {
     "name": "nginx",
     "image": "${nginx_image_identifier}",
+    "memory": 250,
     "essential": true,
     "portMappings": [
       {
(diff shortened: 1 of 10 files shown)


philandstuff:
Thanks to @adityapahuja who did all the brainwork behind this PR, I just plumbed together some terraform

Krenair (approved):

chrisfarms (approved):

adityapahuja (approved):

PR #548, opened 2020-09-11
hub deployer role: deny ability to release elastic IPs

risicle:
https://trello.com/c/lQs5jsln

We don't want this happening accidentally as our IPs tend to be specially
blessed elsewhere.


blairboy362 (dismissed):

Krenair (approved):

Krenair:
Just noticed something - this is a statement in the assume role policy, rather than a policy attached to the role e.g. administrator

blairboy362 (approved):

Krenair (approved):

risicle:
What I've done is an "inline policy", right?

risicle:
Oh and and an `assume_role_policy` is something totally different..

Krenair:
yep, and yes. assume role policies are used to define who can assume a role, what we want to do here is define the permissions of the role itself :)

PR #553, opened 2020-09-23
Add existing "Journeys" dashboard

galund:
- this is a long-standing dashboard that never made it into version control
 - this commit also includes a name change for the ratio of new arrivals to IDP requests
- I've moved it out of the 'sandbox' folder in Grafana which doesn't seem to affect anything but hopefully suggests to people that they ought to make sure version control is up to date.


idrop (approved):
Thanks!

PR #554, opened 2020-09-23
Add existing ECS Journeys dashboard

idrop:
- this is a long-standing dashboard that never made it into version control
 - this commit also includes a name change for the ratio of new arrivals to IDP requests


galund (approved):
Yep, @CharlesIC this is clearly what you were trying to explain to me, but I still don't understand how the same commit can be in the history more than once!

PR #571, opened 2021-01-21
Remove event emitter URL from egress proxy safelist

Krenair:
Based on @instantiator's review of both the code and the event system database,
saml-engine does not submit to the event system, so does not require this
safelist entry.

### DIFF
--- a/terraform/modules/hub/egress_proxy.tf
+++ b/terraform/modules/hub/egress_proxy.tf
@@ -51,14 +51,9 @@ resource "aws_security_group_rule" "egress_proxy_ingress_from_tasks" {
 }
 
 locals {
-  event_emitter_api_gateway = split("/", replace(var.event_emitter_api_gateway_url, "https://", ""))
-}
-
-locals {
   egress_proxy_allowlist_list = [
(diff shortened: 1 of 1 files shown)


instantiator:
Looking further into the code I have found the following:

* `policy` - defines `HubEventLogger`, which contains an `EventSinkProxy`
* `saml-soap-proxy` - defines `ExternalCommunicationEventLogger`, which contains an `EventSinkProxy`  
* `saml-engine` - also defines `ExternalCommunicationEventLogger`, which contains an `EventSinkProxy`

I'm now pretty sure that `saml-engine` and `saml-soap-proxy` _do_ submit events to the event system.

This is corroborated by the values we find in `originatingService` in the event system db audit table: `policy`, `saml-proxy`, `saml-soap-proxy`

I believe we should close this PR - I was mistaken.

Krenair:
Thanks for digging into it Lewis

instantiator (approved):

instantiator:
🤞 for Monday

instantiator (approved):

jkunle (approved):

PR #573, opened 2021-02-03
HUB-814: Add VSP version panel to ECS dashboard

Wynndow:
We've started to send the VSP version used to Prometheus. This adds a
panel to the ECS dashboard to display it.

The easies way to review is probably to extract the JSON model of the current ECS dashboard (which includes the new panel) and compare it to the JSON blob being updated here. From the dashboard click the cog in the top right, then JSON Model.

Screenshot of dashboard including the panel:
Screenshot 2021-02-03 at 10 16 08

### DIFF
--- a/dashboards/Hub-ECS.json
+++ b/dashboards/Hub-ECS.json
@@ -16,7 +16,7 @@
   "gnetId": null,
   "graphTooltip": 1,
   "id": 5,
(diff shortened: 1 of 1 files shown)


CharlesIC (approved):
`VSP versions in the wild`...
Why not 🤷

PR #574, opened 2021-02-04
HUB-813 Update Dockerfiles to use GHCR

j-maynard:


dockerfiles/aws-ruby/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ruby:2.6.1
+FROM ghcr.io/alphagov/verify/ruby:2.6.6
 
 LABEL ruby="2.6.1"
```
galund, line 3: Surely the label should match

galund (changes requested):

dockerfiles/aws-ruby/Dockerfile
CharlesIC: That's the problem with having a vendoring pipeline - it's not great for having specific versions. We are planning to solve this

CharlesIC (approved):
We should check stuff works on updated versions of things

dockerfiles/aws-ruby/Dockerfile
j-maynard: @galund 2.6.6 is a minor point release and doesn't break this.  It also put this on par with all our other repositories which all use Ruby 2.6.6.

galund: In which case the label should just be "2.6" - this makes no sense as is, and will just confuse us.

galund (approved):

PR #576, opened 2021-02-18
HUB-843: Create S3 bucket for hub key rotation

Wynndow:
**See the companion PR on verify-infrastructure-config here: https://github.com/alphagov/verify-infrastructure-config/pull/483**

This adds a module for an S3 bucket for the new hub key rotation
procedure. The module is invoked from the verify-infrastructure-config
repo for each environment.

### DIFF
--- a/terraform/modules/hub-key-rotation/s3.tf
+++ b/terraform/modules/hub-key-rotation/s3.tf
@@ -1,0 +1,8 @@
+resource "aws_s3_bucket" "hub_key_rotation" {
+  bucket = "govukverify-${var.environment}-hub-key-rotation"
+
+  tags = {
+    Environment = var.environment
+    ManagedBy   = "terraform"
+  }
+}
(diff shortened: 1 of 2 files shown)


terraform/modules/hub-key-rotation/s3.tf
```diff
@@ -0,0 +1,8 @@
+resource "aws_s3_bucket" "hub_key_rotation" {
+  bucket = "govukverify-${var.environment}-hub-key-rotation"
```
CharlesIC, line 2: Do we need `${var.environment}` in the bucket name?

CharlesIC: I would also argue we don't need the `govukverify` either 😬

Wynndow: Yep we need it. Bucket names are global, with a region (US, China, rest of the world). The `govukverify` part isn't strictly necessary but is what we do elsewhere and is probably good practices to ensure we're only operating in our own namespace.

CharlesIC: All right then....

CharlesIC (approved):

PR #578, opened 2021-02-26
Switch to ECS Fargate platform version 1.4.0

cbaines:
From the current LATEST, which is 1.3.0. Updating this manually gets
ahead of AWS making this happen on the 8th of March (2021) when they
change the definition of LATEST.

### DIFF
--- a/terraform/modules/hub/analytics.tf
+++ b/terraform/modules/hub/analytics.tf
@@ -75,7 +75,8 @@ resource "aws_ecs_service" "analytics_fargate" {
   deployment_minimum_healthy_percent = 50
   deployment_maximum_percent         = 100
 
-  launch_type = "FARGATE"
+  launch_type      = "FARGATE"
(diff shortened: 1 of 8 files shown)


Krenair:
Do we also want to do metadata/self-service as well as the core hub apps?

cbaines:
> Do we also want to do metadata/self-service as well as the core hub apps?

I don't know if this makes sense in terms of what those things are, but I'll go with yes for now, on the basis on consistency.

I've updated the commit to switch everything I can see using Fargate to the 1.4.0 platform version.

Krenair (dismissed):

Krenair:
@cbaines you didn't sign the commit :(

cbaines:
> @cbaines you didn't sign the commit :(

Ah, I've now pushed a signed commit.

Krenair (approved):

PR #579, opened 2021-03-23
BAU Slack alert for low egress rate to IdPs compared to last month

idrop:
A new Grafana slack alert set on[ Hub ECS Journeys](https://grafana.tools.signin.service.gov.uk/d/UBaTMZJWz/hub-ecs-journeys?orgId=1&var-timerange=1h&from=now-2d&to=now&refresh=1m) which fires when the rate of egress to IdPs is low compared to same time last month.

We can add more alerts into this panel for e.g. low ingress from RPs compared to a month ago.

NB. This panel has already been added, this PR preserves it when Grafana is cycled.

Screenshot 2021-03-23 at 17 19 02

### DIFF
--- a/dashboards/Hub-ECS-Journeys.json
+++ b/dashboards/Hub-ECS-Journeys.json
@@ -16,7 +16,7 @@
   "gnetId": null,
   "graphTooltip": 0,
   "id": 123,
-  "iteration": 1600856137935,
+  "iteration": 1616518012229,
   "links": [],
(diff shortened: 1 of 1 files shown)


JohnWattsGDS (approved):

JohnWattsGDS:
I saw this being born so know what it's supposed to do and hopefully it won't do it when we don't want it to.

PR #580, opened 2021-03-31
BAU Only evaluate the rate of egress alert outside the 0-4am range

idrop:
Dont eval the low egress alert on the [Hub ECS Journeys dash](https://grafana.tools.signin.service.gov.uk/d/UBaTMZJWz/hub-ecs-journeys?orgId=1&var-timerange=1h&from=now-2d&to=now&refresh=1m) from midnight to 4am.
Screenshot 2021-03-31 at 09 37 12
NB. this change has already been made. The PR saves the config in order to restore if Grafana loses its cached version.

* Only evaluate the low egress rate to IDPs outside midnight to 4am, as the measures are too low during this time for a significant comparison.
* Make the anomoly rate 0.8 (from 0.9) which makes the evaluation less sensitive
* Removed the panel for "ratio of requests made from the hub to IDPs vs all requests made from RPs to the hub" as its not really useful
* Tidied up look and feel 

#### Add hidden query B for `hour()`
Screenshot 2021-03-31 at 09 30 29

#### Add outside range clause to alert
Screenshot 2021-03-31 at 09 31 54

### DIFF
--- a/dashboards/Hub-ECS-Journeys.json
+++ b/dashboards/Hub-ECS-Journeys.json
@@ -16,7 +16,7 @@
   "gnetId": null,
   "graphTooltip": 0,
   "id": 123,
-  "iteration": 1616518012229,
+  "iteration": 1617178676236,
   "links": [],
   "panels": [
     {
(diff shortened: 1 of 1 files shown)


Wynndow (approved):
Neat!

PR #589, opened 2021-07-05
[HUB-979] Add aws-terraform Dockerfile

chriscoffee:
In order to run verify-terraform we now 100% need to be using the latest
rekeyed versions of terraform after the codecov.io issue.

Sadly, the aws-terraform containers vendored by autom8 don't really
touch anything below 0.14 (with the correct keys).

We're still using 0.12, and 0.11 in some places so we need to support
these for the time being until we get around to upgrading.

Signed-off-by: Chris Mills <7100370+chriscoffee@users.noreply.github.com>

### DIFF
--- a/dockerfiles/aws-terraform/Dockerfile
+++ b/dockerfiles/aws-terraform/Dockerfile
@@ -1,0 +1,17 @@
+ARG ubuntu_digest
+ARG terraform_digest
+ARG terraform_version
+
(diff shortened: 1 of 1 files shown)


philandstuff:
@Wynndow is this PR still relevant?

Wynndow:
@philandstuff Nope - we did the work to upgrade Terraform to v1+. I'll close this.

PR #600, opened 2021-09-03
HUB-1036: create ecr repo  with push/pull repo policy and renamed hkr tf file

ctazeez:


terraform/modules/hub/hub_key_rotation.tf
```diff
@@ -0,0 +1,103 @@
...
+    ]
+  }
+}
+
+resource "aws_s3_bucket_policy" "hkr_sam_policy" {
+  bucket = aws_s3_bucket.hkr_sam.id
+  policy = data.aws_iam_policy_document.hkr_sam_bucket_policy.json
+}
+
+resource "aws_ecr_repository" "hub_key_rotation" {
```
Wynndow, line 50: What's this repo going to contain? Will there be a repo per account/env, or just one global image?
If the latter, it should probably go with all the other ECR repos [here](https://github.com/alphagov/verify-terraform/blob/master/environments/tools/platform/ecr.tf)

ctazeez: This repo will contain the images created for each  step functions used for hub key rotation. This will be per env and multiple images.

Wynndow: Is each image going to live under a different tag?

Wynndow (approved):

idrop (approved):
👍

PR #601, opened 2021-09-23
HUB-1057 Automate Prometheus Rebooting

chriscoffee:
I've added a few options to the unattended upgrade configuration to automatically reboot these instances at 0300 if there's any updates to apply, it'll also log the output to syslog so we can see it in CloudWatch and track any reboots/issue with reboot needed. I could have done this another way but I thought this option was much simpler for whoever is taking this over to unpick further down the line in an already complex mess of infrastructure code.

Unattended upgrade should pick up the file `/var/run/reboot-required` like we check for in `/usr/bin/instance-reboot-required-metric.sh` and reboot *without confirmation* at 0300. Then when the machine comes back up that'll be no longer there so `/usr/bin/instance-reboot-required-metric.sh` won't set `node_reboot_required` to 1 anymore and therefore won't fire the alert to slack.

If for some reason, the reboot doesn't happen and `/var/run/reboot-required` sits around longer than 24 hours, alertmanager will fire the alert to slack and someone will have to go manually reboot these in concourse!

### DIFF
--- a/terraform/modules/hub/files/cloud-init/prometheus.sh
+++ b/terraform/modules/hub/files/cloud-init/prometheus.sh
@@ -166,7 +166,6 @@ function run-until-success() {
 
 # Make sure unattended upgrades is available and we automatically reboot at 3:00
 # as required in HUB-1057.
-echo unattended-upgrades unattended-upgrades/enable_auto_updates boolean true | debconf-set-selections
 run-until-success "apt-get install --yes unattended-upgrades update-notifier-common"
 cat >> /etc/apt/apt.conf.d/50unattended-upgrades <<'EOF'
 Unattended-Upgrade::Automatic-Reboot "true";
--- a/terraform/modules/hub/files/cloud-init/prometheus.sh
+++ b/terraform/modules/hub/files/cloud-init/prometheus.sh
@@ -164,6 +164,17 @@ function run-until-success() {
   done
 }
 
+# Make sure unattended upgrades is available and we automatically reboot at 3:00
+# as required in HUB-1057.
+echo unattended-upgrades unattended-upgrades/enable_auto_updates boolean true | debconf-set-selections
+run-until-success "apt-get install --yes unattended-upgrades update-notifier-common"
+cat >> /etc/apt/apt.conf.d/50unattended-upgrades <<'EOF'
+Unattended-Upgrade::Automatic-Reboot "true";
+Unattended-Upgrade::Automatic-Reboot-WithUsers "false";
+Unattended-Upgrade::Automatic-Reboot-Time "03:00";
+Unattended-Upgrade::SyslogEnable "true";
+EOF
+
 # ECS
 echo 'Running ECS using Docker'
 mkdir -p /etc/ecs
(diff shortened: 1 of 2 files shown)


Wynndow:
This means we'll be rebooting all envs at the same time, so lose the advantage of running staging/integration before prod to highlight any issues with the new upgrades. How risky is that?

Wynndow:
I've jumped on to a staging prom instance and applied these changes manually (apart from the alerts config) and it seems to do what it's supposed to do. There wasn't anything that needed a reboot though so will have to make sure it'd done the right thing when an upgrade is availabe.

chriscoffee:
> This means we'll be rebooting all envs at the same time, so lose the advantage of running staging/integration before prod to highlight any issues with the new upgrades. How risky is that?
> 
> Edit: Also, thinking about this, we run run three ECS prometheus services, each with one task. If we make this change we'll shut them all down at exactly the same time and lose all prometheus coverage for a while. Might it be better to orchestrate doing them service by service?

So yes this will reboot all of them at the same time, and not just on an environment level, it'll reboot the nodes at the same time too.

They're only really security updates so it shouldn't cause too much hassle. However there is always the risk that something could go wrong with the updated packages and dep changes.

The other option is we hook reboot-required into the concourse reboot script which felt hacky and overly complex for just simple restarts at a specific time.

However, I think if the risk is something we see as not worth it I can have a rethink to be honest. I think I'll ponder an alternative over lunch

CharlesIC (approved):
This is scheduled for 3am so I think this is fine as we don't get much usage at that time

terraform/modules/hub/files/cloud-init/prometheus.sh
```diff
@@ -164,6 +164,16 @@ function run-until-success() {
   done
 }
 
+# Make sure unattended upgrades is available and we automatically reboot at 3:00
+# as required in HUB-1057.
+run-until-success "apt-get install --yes unattended-upgrades update-notifier-common"
+cat >> /etc/apt/apt.conf.d/50unattended-upgrades <<'EOF'
+Unattended-Upgrade::Automatic-Reboot "true";
+Unattended-Upgrade::Automatic-Reboot-WithUsers "false";
+Unattended-Upgrade::Automatic-Reboot-Time "03:00";
```
CharlesIC, line 173: Is this too crazy?

```suggestion
Unattended-Upgrade::Automatic-Reboot-Time "03:$((10 + $RANDOM % 60))";
```

Set minutes to a random number between 10 and 59 to spread out the restarts and reduce the chances of losing Prometheus coverage.

Wynndow: It still leaves the issue of upgrading prod before staging/integration. Better to bork staging and not deploy to prod than just deploying blind.

CharlesIC: Yeah fair enough, it looks like we'd need a Concourse job for that then

simongurney1: done with https://github.com/alphagov/verify-infrastructure-config/pull/602

PR #603, opened 2021-09-29
HUB-1067: attach ingress and egress rule to lambda security group to talk to internal lb

ctazeez:


terraform/modules/hub/hub_key_rotation.tf
```diff
@@ -116,6 +116,24 @@ resource "aws_security_group" "sg_hkr_lambda" {
   vpc_id = aws_vpc.hub.id
 }
 
+resource "aws_security_group_rule" "sg_hkr_lambda_ingress" {
```
simongurney1, line 119: I dont think ingress is required is it?

simongurney1: Ah is this ingress from the lambda?  If so, is it not just port 80?

ctazeez: added incase it required any other resource need to talk to lambda since its attached to vpc

ctazeez: its iingress for the lambda

simongurney1: ah, in that case I think it should be removed unless we actually need it

terraform/modules/hub/hub_key_rotation.tf
```diff
@@ -116,6 +116,24 @@ resource "aws_security_group" "sg_hkr_lambda" {
...
+  type      = "ingress"
+  from_port = "0"
+  to_port   = "65535"
+  protocol  = "tcp"
+  cidr_blocks = [aws_vpc.hub.cidr_block]
+  security_group_id = aws_security_group.sg_hkr_lambda.id
+}
+
+resource "aws_security_group_rule" "sg_hkr_lambda_egress" {
```
simongurney1, line 128: I don't normally like open egress, but this is the behaviour of the existing lambdas so im fine with it

terraform/modules/hub/hub_key_rotation.tf
ctazeez: Done

simongurney1 (approved):

PR #611, opened 2022-03-28
BAU: Increase prometheus volume size

Wynndow:
One of the prometheus ECS services is failing to start up with what
looks like an issue with disk space:

`err="persist head block: write compaction: write chunks: no space left on device"`

The prometheus services have a 100GB volume mapped into them that
prometheus writes to. This volume for prom-1 is basically full:

`/dev/nvme1n1   ext4       98G   93G  345M 100% /srv/prometheus.`

The other two volumes are not far behind.

This PR increases the size of the volume to fix the issue in the short
term. We probably want to do something better though to stop this issue
re-occurring.

I've planned this change against staging and can see the volume us
updated in place rather than replaced which means we won't lose the data
on it.

### DIFF
--- a/terraform/modules/hub/variables.tf
+++ b/terraform/modules/hub/variables.tf
@@ -75,7 +75,7 @@ variable "saml_proxy_memory_limit_mb" {
 }
 
 variable "prometheus_volume_size" {
-  default = 100
+  default = 120
 }
 
 variable "publically_accessible_from_cidrs" {


simongurney1 (approved):
Have we got to do something inside the ec2 instance to tell it to extend the partition?

Wynndow:
@simongurney1 We've got this in the init script: https://github.com/alphagov/verify-infrastructure/blob/master/terraform/modules/hub/files/cloud-init/prometheus.sh#L99

Thinking about it, it _probably_ won't increase it. But if we run this and it doesn't, AWS document the process pretty well so we can do it manually: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/recognize-expanded-volume-linux.html

PR #614, opened 2022-07-07
fix: change interval

simongurney1:


terraform/modules/hub/prometheus.tf
```diff
@@ -393,13 +393,20 @@ resource "aws_dlm_lifecycle_policy" "prometheus" {
     resource_types = ["VOLUME"]
 
     schedule {
-      name = "24 hours weeks of hourly snapshots"
-      create_rule { interval = "cron(0 * * * ? *)" }
+      name = "24 hourly snapshots"
```
andylockran, line 396: I'm slightly confused by this description, can we make it more explicit to describe the interval and the retention window.

andylockran, line 396:
```suggestion
      name = "hourly snapshots, retained for 24 hours."
```

terraform/modules/hub/prometheus.tf
```diff
@@ -393,13 +393,20 @@ resource "aws_dlm_lifecycle_policy" "prometheus" {
...
+      create_rule { 
+        interval = 1 
+        interval_unit = "HOURS"
+        }
       retain_rule { count = 24 }
     }
     schedule {
       name = "2 weeks of daily snapshots"
```
andylockran, line 404:
```suggestion
      name = "daily snapshots, retained for 2 weeks"
```

jimmc400 (approved):

PR #627, opened 2022-12-13
BAU - add www-data user

dushan-madetech:


owenblacker (changes requested):
That will error. I changed it from user `www-data` to user `nginx` to prevent it erroring, if you look in the history

owenblacker (approved):
Let's try it anyway; we can revert if required

